Issue #27, opened 2024-01-01
API: Include unevolved Pokémon in /pokemon response


SleepAPI:
### Details

This is blocked by: https://github.com/SleepAPI/SleepAPI/issues/5

The /pokemon API route is used to return a list of all available Pokémon. 

Make sure that we use the new POKEDEX constant with both optimal and unevolved Pokemon and that this route does not only return the optimal list. This has to be checked after #5 has been implemented. 

 This can be done directly in the `getPokemonNames` helper function in `pokemon-utils.ts`. 

Update the tests too.

PR #42, opened 2024-01-06
Use complete pokedex for production calculator

3geek14:
Add all Pokemon to production calculator

Also alphebetize the list, and use the in-game names for event Pokemon.


backend/src/utils/pokemon-utils/pokemon-utils.ts
SleepAPI, line 1: I'm not sure what prompted this change, while it's fine, it's also not needed for this PR.
Were you trying to solve #27 at the same time perhaps? If so, that would be the function below `getPokemonNames`. Then you would need to update the tests too and while you're at it getPokemonForName doesn't appear to be covered by tests.

I'm also fine with doing this in a different PR though, just mentioned it as I was unsure what prompted this change.

SleepAPI (commented):
Let one comment to check out, but everything looks great to me!
I'll approve after backend is merged, just to see that this PR is updated correctly accordingly

3geek14:
Removed the extra file and the association with #27. Tested locally that the production calculator shows added Pokemon in the dropdown.

SleepAPI (approved):
Looks great to me! Wonderful contribution!

PR #47, opened 2024-01-07
build(deps): updated unused and stale deps

SleepAPI:
**WHY**

The dependabot found some vulnerabilities it could not resolve. This fixes those.

**WHAT**

We had some unused dependency artifacts that are removed here. Also updated express minor to fix security vulnerability.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -9,18 +9,14 @@
       "version": "5.0.0",
       "dependencies": {
         "@sinclair/typebox": "^0.31.22",
-        "@types/ejs": "^3.1.4",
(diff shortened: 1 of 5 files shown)


3geek14:
Should probably leave this open. The PR had a scope on the have, but not the commit.

SleepAPI:
> Should probably leave this open. The PR had a scope on the have, but not the commit.

I'm not sure what you mean, the commit has a scope in the commit header

3geek14:
Oh, I was reading the wrong thing and left the comment in the wrong place. Morning brain, sorry.

PR #49, opened 2024-01-07
Update documentation for Windows development

sethdw:
I wrote down a couple of things as I was setting up the development environment on my Windows machine.

Let me know if my commits are too small - we're quite zealous with "only one change per commit" at my work!

### DIFF
--- a/DEVELOPMENT_SETUP.md
+++ b/DEVELOPMENT_SETUP.md
@@ -2,19 +2,20 @@
 
 The project acts as a monorepo where each module is individually deployed. It's written in Node.js with Typescript.
 
-| Application | Description|
-| - | -
-| backend | This deployment exposes the public API: [Sleep API][sleepapi]. This is where the simulations, number-crunching and rankings happen. Currently this also exposes a basic html/js/css website, but this is due for replacement.
-| frontend | A work-in-progress website. The idea is to deploy a full-fledged proper website with Typescript + Vue + Node as a static web app in Azure.
-| bot | A work-in-progress Discord bot. Currently this just sets up a basic ping command, but it is deployed and hosted 24/7. There are plans to expand this with functionality from Sleep API.
+| Application | Description                                                                                                                                                                                                                   |
+| ----------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


DEVELOPMENT_SETUP.md
```diff
@@ -2,19 +2,26 @@
...
-| Tool   | URL                                               |
-| ------ | ------------------------------------------------- |
-| VSCode | <https://code.visualstudio.com/>                  |
-| NVM    | <https://github.com/nvm-sh/nvm>                   |
+| Tool   | URL                              |
+| ------ | -------------------------------- |
+| VSCode | <https://code.visualstudio.com/> |
+| NVM    | <https://github.com/nvm-sh/nvm>  |
+
+On Windows it is recommended to use nvm inside WSL (Windows Subsystem for Linux) rather than nvm-windows or other alternatives.
```
SleepAPI, line 20: Have you tried `nvm-windows`? I have some concern that recommending setting up WSL in order to develop for Sleep API on Windows might be a tough barrier of entry for users without WSL.

SleepAPI (commented):
Just waiting for decision regarding squash vs rebase merge in discord dev channel before approve. Left a comment regarding nvm on Windows. Looks good to me beyond that. Great stuff, thanks for doing this 👍

DEVELOPMENT_SETUP.md
sethdw: I haven't, but honestly I still wouldn't recommend it. Once WSL is set up it really does "just work" and makes other setup far easier - especially the Docker setup later.

SleepAPI: Perhaps we could phrase it as "it is recommend to use nvm insice WSL, but other alternatives like nvm-windows exist. Current phrasing makes it seem like nvm-windows is not a good option which I'm not sure of. If people are turned off by the WSL then perhaps they should give nvm-windows a shot rather than giving up.

After that I would personally prefer slightly bigger commits with the reasoning being that it makes the develop/main branch history more parseable if we can just scroll through some slogans of feature commits. Small commits are great for reviewing though. I'm open to suggestions and discussion as always.

SleepAPI (approved):

PR #65, opened 2024-01-11
feat: custom error file

AdeosunDeji:


backend/src/domain/error.ts
SleepAPI, line 1: I see you've done your research in the code, excellent! Some great suggestions in here.

Regarding the structure I could have been more explicit in the ticket. I would like the custom errors to extend the Error base class.

Furthermore we should create a folder called error under domain. Then move `error.ts` into `domain/error/error.ts`. 

Perhaps we should even rename `error.ts` to `sleepapi-error.ts`and add in a base Sleep API error class like:
```typescript
export class SleepAPIError extends Error {
  static message(message: string) {
    return new this(message);
  }
}
```

After that we can create folders under `domain/error` to group different error types. For example we could have `domain/error/database-error` and move `DatabaseMigrationError`, `DatabaseConnectionError` etc in there.

Instead of using const variables for the errors I would like them to extend this `SleepAPIError` base error class.
So it would look something like:
```typescript
export class DatabaseMigrationError extends SleepAPIError {}
```

This should allow us to re-use the error types without limiting which message we include as the base `SleepAPIError` would allow for passing custom messages.

For example `DatabaseMigrationError` could be used both for incorrect migration type as you suggest right now, or if we crash during migration due to a schema issue etc.

This also means we're mostly looking to setup the types here rather than the messages. The message we'll pass just like we're passing them to the normal `Error` base class as we do today.

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
+export const DatabaseMigrationError = 'DATABASE_MIGRATION is optional, but if set must be one of [UP, DOWN]';
+export const DatabaseConnectionError = 'Missing environment variables for database connection';
+export const EntryNotFoundError = (tableName: string, filter: any) =>
```
SleepAPI, line 3: To align with your other naming standard suggestions I would rename the following database errors:
`EntryNotFoundError` to `DatabaseNotFoundError`
`InsertError` to `DatabaseInsertError`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
+export const DatabaseMigrationError = 'DATABASE_MIGRATION is optional, but if set must be one of [UP, DOWN]';
+export const DatabaseConnectionError = 'Missing environment variables for database connection';
+export const EntryNotFoundError = (tableName: string, filter: any) =>
+  `Unable to find entry in ${tableName} with filter [${JSON.stringify(filter)}]`;
+export const InsertError = (resultLength: number) => `Insert expected one element but was ${resultLength}`;
+export const MealNotFoundError = (mealName: string) => `Couldn't find meal with name:  ${mealName.toUpperCase()})`;
```
SleepAPI, line 6: As per my previous suggestion of creating common re-usable error types this would not be needed.
Instead we'll throw a `DatabaseNotFoundError` and pass in the meal-specific message.

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
+export const DatabaseMigrationError = 'DATABASE_MIGRATION is optional, but if set must be one of [UP, DOWN]';
+export const DatabaseConnectionError = 'Missing environment variables for database connection';
+export const EntryNotFoundError = (tableName: string, filter: any) =>
+  `Unable to find entry in ${tableName} with filter [${JSON.stringify(filter)}]`;
+export const InsertError = (resultLength: number) => `Insert expected one element but was ${resultLength}`;
+export const MealNotFoundError = (mealName: string) => `Couldn't find meal with name:  ${mealName.toUpperCase()})`;
+export const Level60Error = 'Somehow level 60 produced ingredients not populated for buddy-combination-dao';
```
SleepAPI, line 7: Regarding the following errors:
- `Level60Error`
- `EnergyPercentageError`
- `LevelError`
- `NatureError`

These could all be replaced by an error called for example `StatError` and be moved into `domain/error/stat/stat-error.ts`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+export const EntryNotFoundError = (tableName: string, filter: any) =>
+  `Unable to find entry in ${tableName} with filter [${JSON.stringify(filter)}]`;
+export const InsertError = (resultLength: number) => `Insert expected one element but was ${resultLength}`;
+export const MealNotFoundError = (mealName: string) => `Couldn't find meal with name:  ${mealName.toUpperCase()})`;
+export const Level60Error = 'Somehow level 60 produced ingredients not populated for buddy-combination-dao';
+export const EnergyPercentageError = (energyPercentage: number) =>
+  `Energy has to be between 0 and 150, but was ${energyPercentage}`;
+export const LevelError = 'Only level 1-100 allowed';
+export const NatureError = (natureName: string) => `Couldn't find nature with name:  ${natureName.toUpperCase()}`;
+export const PokemonNameError = (pokemonName: string) =>
```
SleepAPI, line 12: This could be renamed to just `PokemonError` and be moved into `domain/error/pokemon/pokemon-error.ts`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+  `Couldn't find nature with name:  ${pokemonName.toUpperCase()}`;
+export const SubTeamsError = 'Should not happen since we check maxTeamSize > 1, and we return only undefined for 0';
```
SleepAPI, line 14: This `SubTeamsError`can be removed since this indicates a ProgrammingError which you already have added

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+export const IngredientError = (name: string) => `Ingredient with name [${name}] does not exist`;
```
SleepAPI, line 15: This could be moved into `domain/error/ingredient/ingredient-error.ts`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+export const ProgrammerError = (name: string) =>
```
SleepAPI, line 16: I would like to rename `ProgrammerError` to `ProgrammingError` as it suggest an issue with the code rather than blaming an individual programmer. This can be moved into `domain/error/programming/programming-error.ts`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+  `Programmer error: Can't find Pokemon coming from DB with name ${name}`;
+export const ObjectError = 'Only one object may be mocked per invocation to mock';
```
SleepAPI, line 18: We should rename this to `MockError` and perhaps we can create a test folder under domain for this so we may add in more test-related errors. So move the `MockError` into `domain/error/test/mock-error.ts`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+export const MissingEnvError = 'Missing environment variables';
```
SleepAPI, line 19: How about something more generic like `ConfigError` and move it into `domain/error/config/config-error.ts`

backend/src/domain/error.ts
```diff
@@ -0,0 +1,20 @@
...
+export const FindMealError = (name: string) => `Couldn't find meal with name:  ${name.toUpperCase()})`;
```
SleepAPI, line 20: This could be renamed to `MealError` and be moved into `domain/error/meal/meal-error.ts`

SleepAPI (changes requested):
It looks like your files use CRLF line ending so I'm assuming you're developing on a Windows machine. This causes big diffs for all files that you saved on your machine.
This also makes the PR difficult to review and if we allowed formatting changes between different platforms then every other PR would have this issue.

We have explicitly added `"files.eol": "\n"` to the `sleepapi.code-workspace` to fix this.
So this suggests to me that likely you either didn't work in Visual Studio Code or didn't use the workspace. Please read our [DEVELOPMENT_SETUP](https://github.com/SleepAPI/SleepAPI/blob/develop/DEVELOPMENT_SETUP.md) if you need help with this.

Furthermore it seems that you removed some line endings. Prettier is set as the default autoformat in the workspace and prettier is configured by default to add newlines at the end of the file. Prettier also formats LF as line endings by default. This also suggests to me that VS Code and/or the workspace was not used.

I noticed our .editorconfig states we should not add new lines at the end of a file, but this is unintended and as previously mentioned gets overridden by prettier.

Final note; great suggestions for different error types. You've clearly gone through the code and summarized which errors we throw today, excellent work! 

I left some comments on how I think we should structure the errors. 
Please fix those and the LF/newline formatting 😃

backend/src/domain/error.ts
AdeosunDeji: Thanks for the corrections and guidance. Your insight is greatly appreciated.

AdeosunDeji (commented):
Made changes. Waiting for your feedback.

SleepAPI (changes requested):
Big improvement. Left a few comments and I have a few more comments here.

Currently you implement each error class like so:
```typescript
export class DatabaseMigrationError extends SleepAPIError {
  constructor(message = 'DATABASE_MIGRATION is optional, but if set must be one of [UP, DOWN]') {
    super(message);
    this.name = 'DatabaseMigrationError';
  }
}
```

I would prefer it to be completely open types instead like:
```typescript
export class DatabaseMigrationError extends SleepAPIError {}
```

That way you can throw the same error in many places, but pass different messages, making it more generic. I would like to avoid case-specific message in places where we're trying to be generic. Please update all error types to open objects.

It also looks like you're still writing the files with CRLF instead of LF. Please see my comment on the previous commit detailing how to correctly format the files with VS Code and the workspace.

Finally I see you're adding new commits to the PR, that's perfectly fine, but since we're just fixing the commits introduced on this PR I would like the final result of them to be just one commit. So either you can squash those 2 commits now and then amend any future fixes or keep pushing new commits and then in the end squash them all together before I merge.

Please let me know if you have any questions 👍

PR #71, opened 2024-01-12
feat: custom error

AdeosunDeji:


AdeosunDeji (commented):
I opened with sleepapi.code-workspace. Vs code says i'm writing with LF. I hope i haven't pushed something wrong this time. Awaiting your feedback

SleepAPI (changes requested):
Unfortunately the PR still includes a lot of files with CRLF reformatting.
You can see the "Files changed" as [138 different files](https://github.com/SleepAPI/SleepAPI/pull/71/files) on the PR, but you're just trying to change like 9 files. 

Alright, but here's for the good news! It looks like for the new files that you changed (the error type files) you are now indeed using LF formatting correctly. So the workspace works! I guess you just have to re-save the files you previously reformatted to CRLF so they can be formatted back to LF.

You can see which files you reformatted on the [Files changed](https://github.com/SleepAPI/SleepAPI/pull/71/files) section on the PR. Any file that is changed that is not an error file only seems to have formatting changes.

SleepAPI (changes requested):
You're almost there! 

A few files left under backend that need to be reformatted back to LF:
- .eslintrc.cjs
- .prettierignore
- .prettierrc.json
- package-lock.json

I also understand why you didn't reformat these back. The reason is that the workspace file by default hides these from showing up, as they're mostly uninteresting for development. Now that you're using the workspace these don't show up anymore.

In the workspace file you'll find:
```json
"files.exclude": {
  "backend": true,
  "frontend": true,
  "bot": true,
  ".vscode": true,
  "dist": true,
  "node_modules": true,
  "package-lock.json": true,
  "*.tsbuildinfo": true,
  "coverage": true,
  ".eslintrc.cjs": true,
  ".prettierignore": true
},
```

You can *temporarily* remove this and save. That will cause the files to show up. You can then reformat them. Don't forget to add back the files.exclude to the workspace after.

After you've fixed that please either squash the commits into one commit and force push the changes to the PR or create a new PR like you did before. The goal is to only have one commit on the PR with the error types added.

Good work so far! We're almost there 👍

PR #81, opened 2024-01-15
implement custom errors

AdeosunDeji:


SleepAPI:
The commit doesn't seem entirely correct. It looks like you squashed the last two commits on your fork.
That would be the first commit you made, which we merged to develop last time and then this new change.

Then you opened a pull request for that squashed commit which means the pull request now has a commit that says it's authored by both of us, since you squashed the commit that I merged last, but this commit is all you. 
What also now happens is the commit on this PR includes all the changes you already made and merged last time.

You might need to re-fork or reset your fork to the latest develop. Sleep API develop has received some commits since you last contributed so your fork is also out of date by a few commits.

backend/src/utils/pokemon-utils/pokemon-utils.ts
```diff
@@ -1,10 +1,11 @@
+import { ProgrammingError } from '../../domain/error/programming/programming-error';
 import { OPTIMAL_POKEDEX, Pokemon } from '../../domain/pokemon/pokemon';
 import { getBerriesForFilter } from '../berry-utils/berry-utils';
 
 export function getPokemonForName(name: string): Pokemon {
   const pokemon = OPTIMAL_POKEDEX.find((pokemon) => pokemon.name === name);
   if (!pokemon) {
-    throw new Error(`Programmer error: Can't find Pokemon coming from DB with name ${name}`);
+    throw new ProgrammingError(`Programmer error: Can't find Pokemon coming from DB with name ${name}`);
```
SleepAPI, line 8: In this specific case we don't have to mention twice that it's a programmer error, so we can just throw:
```typescript
throw new ProgrammingError(`Can't find Pokemon coming from DB with name ${name}`);
```

backend/src/domain/error/route/route-error.ts
SleepAPI, line 1: Once you update your fork you'll notice that the `queryAsMandatoryNumber` function in `routing-utils.ts` has been removed and thus this RouteError is also not needed anymore. I believe that was the only place you were throwing RouteError.

SleepAPI (changes requested):
Left a few small comments regarding the error replacement, but for the most part it looks great. As mentioned by my previous comment there seems to be some squash issue with the commit itself and also the fork is out-of-date.

PR #82, opened 2024-01-15
feat: implementing custom errors

AdeosunDeji:


AdeosunDeji:
Looks like something is wrong. I don't know what is

SleepAPI (approved):
Looks great! 

The reason the build complained was because the errors you replaced was mainly for untested code. Therefore it complains that we shouldn't add new untested code. But seeing as this was already not tested and you just replaced the error type thrown I won't require you to implement new tests, that's technical debt outside this task.

Thanks for this, this is a great contribution and you grasped the task perfectly @AdeosunDeji

AdeosunDeji:
Thanks for your guidance

Issue #26, opened 2023-12-30
API: Add a default 404 page


SleepAPI:
### Details

Instead of just listing "Cannot GET X" we should add a default 404 page.
It could link back to the root page and mention that you can navigate to other pages from there.

PR #184, opened 2024-02-24
feat: add a 404 page

Schanihbg:
Fix for #26 #129

### DIFF
--- a/backend/src/assets/404.html
+++ b/backend/src/assets/404.html
@@ -1,0 +1,75 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Page not found | Sleep API</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="icon" href="/cook.ico" type="image/x-icon" />
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/css/bootstrap.min.css"
+      integrity="sha384-T3c6CoIi6uLrA9TneNEoa7RxnatzjcDSCmG1MXxSR1GAsXEV/Dwwykc2MPK8M2HN"
+      crossorigin="anonymous"
+    />
+    <link rel="stylesheet" href="/global.css" />
+    <style>
+      #bannerImage {
+        max-width: 90%;
+        height: auto;
+        margin-left: 5%;
+        margin-right: 5%;
+      }
+
+      @media (min-width: 1000px) {
+        .card {
+          width: 80%;
+        }
+
+        #sneaselgif {
+          max-width: 50%;
+        }
+      }
+    </style>
+  </head>
+
+  <body>
+    <div class="container" style="margin-bottom: 50px">
+      <header class="d-flex justify-content-center mb-2 mt-2">
+        <a href="/">
+          <img
+            id="bannerImage"
+            src="sleepapi-banner.png"
+            alt="Sleep API Cooking Simulator Logo"
+            title="Pokémon Sleep Cooking Simulator"
+          />
+        </a>
+      </header>
+      <div class="card-container">
+        <div class="card" style="margin: auto">
+          <div class="card-body">
+            <div class="row justify-content-center mb-2">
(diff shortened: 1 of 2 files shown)


backend/src/assets/404.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Sleep API</title>
```
SleepAPI, line 4: The title should probably be something like "404 Not Found"

backend/src/assets/404.html
```diff
@@ -0,0 +1,76 @@
...
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta
```
SleepAPI, line 7: I don't think we should have a description meta for this, nor should we have open graph headers. We don't want to risk sharing this on social media and baiting uses with a link that's broken.

backend/src/assets/404.html
```diff
@@ -0,0 +1,76 @@
...
+            alt="Sleep API Cooking Simulator Logo"
+            title="Pokémon Sleep Cooking Simulator"
+          />
+        </a>
+      </header>
+      <div class="card-container">
+        <div class="card" style="margin: auto">
+          <div class="card-body">
+            <div class="row justify-content-center mb-2">
+              <h1 style="text-align: center; color: #ddd">Error 404</h1>
```
SleepAPI, line 52: I know I specifically requested ddd, but I think it probably looks better with white in this case since its h1 and logo text color above is also white

SleepAPI: we can leave the "To front page" text color ddd, that looks good

backend/src/assets/404.html
```diff
@@ -0,0 +1,76 @@
...
+                <a href="/" class="btn btn-danger btn-navigation">To front page</a>
+              </h2>
+              <img src="cook.gif" alt="Sneasel cooking" />
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+
+    <footer class="text-center py-2" style="background-color: #555; color: #ddd">
```
SleepAPI, line 64: Footer doesn't stick to bottom of the page, the reason is that app.css isn't imported here.
We kind of want to move away from app.css anyway, it was just a shitty w3schools initial solution like half a year ago. 

Instead you could probably just add this to global.css (which is already imported here):
```css
footer {
  position: absolute;
  left: 0;
  bottom: 0;
  width: 100%;
  background-color: black;
  color: white;
  text-align: center;
}
```

backend/src/assets/404.html
```diff
@@ -0,0 +1,76 @@
...
+      <div class="card-container">
+        <div class="card" style="margin: auto">
+          <div class="card-body">
+            <div class="row justify-content-center mb-2">
+              <h1 style="text-align: center; color: #ddd">Error 404</h1>
+              <h1 style="text-align: center; color: #ddd">Sneasel could not find your page</h1>
+              <h2 style="text-align: center; color: #ddd">
+                <a href="/" class="btn btn-danger btn-navigation">To front page</a>
+              </h2>
+              <img src="cook.gif" alt="Sneasel cooking" />
```
SleepAPI, line 57: This looks great on mobile, but on desktop the image is too big and the container card is probably too wide.
I think we could add id=sneaselgif to the image and add desktop-specific styling to script tag css above like:
```css
@media (min-width: 1000px) {
        .card {
          width: 80%;
        }

        #sneaselgif {
          max-width: 50%;
        }
      }
```

SleepAPI (changes requested):

backend/src/assets/404.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Page not found | Sleep API</title>
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta property="og:url" content="https://sleepapi.net" />
```
SleepAPI, line 7: Can you remove this one too, I don't want any thumbnail for 404 pages and I think this would still cause a barebones thumbnail to appear on social media shares

PR #288, opened 2024-05-17
Add per-ingredient ranking to API

3geek14:
Compute a ranking for individual ingredients. I barely looked at how I want the data to output, so that'll need some attention eventually, but probably not a big deal until it's added to a frontend, either current or new.

Note that this approach doesn't report any mon that finds an ingredient rarely enough to be less than 1 per meal window. So Jolly level 30 Slowking might be excluded, as it finds 0.9 Tails per meal window. Not sure if there are any other settings where Pokemon can be excluded. (Slowpoke also gets <1 Tail, but it's already excluded because it's not fully-evolved.)

### DIFF
--- a/backend/src/services/api-service/optimal/optimal-service.test.ts
+++ b/backend/src/services/api-service/optimal/optimal-service.test.ts
@@ -1,7 +1,7 @@
 import { SetCoverProductionStats } from '@src/domain/computed/production';
-import { berry, dessert, nature, subskill } from 'sleepapi-common';
+import { berry, dessert, ingredient, nature, subskill } from 'sleepapi-common';
 import { prettifyIngredientDrop } from '../../../utils/json/json-utils';
-import { findOptimalSetsForMeal, getOptimalFlexiblePokemon } from './optimal-service';
+import { findOptimalMonsForIngredient, findOptimalSetsForMeal, getOptimalFlexiblePokemon } from './optimal-service';
 
 describe('findOptimalSetsForMeal', () => {
   it('shall find all optimal solutions for a recipe', () => {
@@ -50,6 +50,38 @@ describe('findOptimalSetsForMeal', () => {
         },
       ]
     `);
+  });
+});
+
+describe('findOptimalMonsForIngredient', () => {
+  it('shall find all optimal mons for an ingredient', () => {
+    const input: SetCoverProductionStats = {
+      level: 30,
+      nature: nature.RASH,
+      subskills: [],
+      berries: berry.BERRIES,
+      e4eProcs: 0,
+      e4eLevel: 6,
+      cheer: 0,
+      extraHelpful: 0,
+      helperBoostProcs: 0,
+      helperBoostUnique: 1,
+      helperBoostLevel: 6,
+      helpingBonus: 0,
+      camp: false,
+      erb: 0,
+      incense: false,
+      skillLevel: 6,
+      mainBedtime: { hour: 21, minute: 30, second: 0 },
+      mainWakeup: { hour: 6, minute: 0, second: 0 },
+    };
+    const teams = findOptimalMonsForIngredient(ingredient.SLOWPOKE_TAIL.name, input, 1).teams.filter(
+      (team) => team.surplus.relevant[0].amount >= 0.999
+    );
+    expect(teams).toHaveLength(2);
+    const pokemonNames = teams.map((team) => team.team.map((member) => member.pokemonCombination.pokemon.name));
+    expect(pokemonNames).toContainEqual(['SLOWBRO']);
+    expect(pokemonNames).toContainEqual(['SLOWKING']);
   });
 });
 
--- a/backend/src/services/api-service/optimal/optimal-service.ts
+++ b/backend/src/services/api-service/optimal/optimal-service.ts
@@ -8,6 +8,7 @@ import {
   calculateOptimalProductionForSetCover,
   calculateSetCover,
 } from '@src/services/calculator/set-cover/calculate-set-cover';
+import { getIngredientForName } from '@src/utils/ingredient-utils/ingredient-utils';
 import { calculateCritMultiplier, CritInfo, getMeal, getMealsForFilter } from '@src/utils/meal-utils/meal-utils';
 import {
   calculateCombinedContributions,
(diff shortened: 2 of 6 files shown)


backend/src/services/calculator/per-ingredient-ranker/per-ingredient-ranker.ts
SleepAPI, line 1: I think we can probably remove this file, move the set cover call directly into your optimal-service function which currently is kinda barebones. This one ends up not doing a lot on its own, it mostly just passes on to set-cover

backend/src/services/api-service/optimal/optimal-service.ts
SleepAPI, line 1: Missing testing for new function

backend/src/services/api-service/optimal/optimal-service.ts
```diff
@@ -28,6 +31,34 @@ export function findOptimalSetsForMeal(mealName: string, input: SetCoverProducti
...
+ * Finds the optimal Pokemon for a specific ingredient
+ *
+ * API: /api/optimal/ingredient
+ */
+export function findOptimalMonsForIngredient(
+  ingredientName: string,
+  input: SetCoverProductionStats,
+  monteCarloIterations: number
+) {
+  const ingredient: Ingredient = getIngredientForName(ingredientName);
```
SleepAPI, line 44: minor: but no need to explicitly state return type of Ingredient. The compiler will understand it's Ingredient. The discussion regarding using explicit types vs not is a whole topic on its own, but I prioritize matching overall code structure right now

backend/src/services/api-service/optimal/optimal-service.ts
```diff
@@ -28,6 +31,34 @@ export function findOptimalSetsForMeal(mealName: string, input: SetCoverProducti
...
+
+  const optimalCombinations = rankByIngredient({
+    ingredient,
+    reverseIndex,
+    cache: new Map(),
+    timeout: TEAMFINDER_SET_COVER_TIMEOUT,
+  });
+
+  return {
+    ingredient: ingredient.longName,
```
SleepAPI, line 56: In other parts of the site we currently use the short name more than the long name. You can check the current results on the site, they'll all list "potato" instead of "soft potato". 

I'll leave this decision to you though, but you're currently using the short name as input param in the route too like /api/optimal/ingredient/potato

backend/src/services/calculator/per-ingredient-ranker/per-ingredient-ranker.ts
```diff
@@ -0,0 +1,36 @@
...
+import { IngredientSet } from 'sleepapi-common';
+import { Ingredient } from 'sleepapi-common/dist/domain/ingredient';
+
+export function rankByIngredient(params: {
+  ingredient: Ingredient;
+  cache: Map<string, CustomPokemonCombinationWithProduce[][]>;
+  reverseIndex: Map<string, CustomPokemonCombinationWithProduce[]>;
+  timeout?: number;
+}): OptimalTeamSolution[] {
+  const singletonIngredient: IngredientSet[] = [{ ingredient: params.ingredient, amount: 1 }];
```
SleepAPI, line 29: very minor: but I think singleton generally represents a single instanced and re-used object of a class. In this case it's just a recipe with 1 quantity used as parameter. Perhaps we should call it "singleIngredientRecipe" or something

SleepAPI (changes requested):

backend/src/services/api-service/optimal/optimal-service.ts
SleepAPI: The other comments are quite minor, but this one is important to me

SleepAPI:
Also missing updated Swagger documentation, it's generated automatically on `npm run build`, start or dev, but dev doesn't keep it updated on auto-save which is probably why it's missing here.

backend/src/services/calculator/per-ingredient-ranker/per-ingredient-ranker.ts
3geek14: Done.

3geek14: Done.

backend/src/services/api-service/optimal/optimal-service.ts
3geek14: Added a test. It uses Monte Carlo without a fixed seed, so I wasn't able to assert how many tails each mon finds, but I asserted the number of mons that it finds and which ones they are. Lemme know if you'd like me to change this to make it deterministic?

backend/src/services/api-service/optimal/optimal-service.ts
3geek14: Removed the extra annotations. I'm reminded of the "technical details" for [DreamBerd's type annotations](https://github.com/TodePond/DreamBerd?tab=readme-ov-file#types).

backend/src/services/api-service/optimal/optimal-service.ts
3geek14: Changed to be consistent. In the future, I think "Fancy Apple Curry" and "Soft Potato" should be used in output rather than "FANCY_APPLE_CURRY" and "POTATO", but those changes can be made together, probably after the frontend redesign.

backend/src/services/calculator/per-ingredient-ranker/per-ingredient-ranker.ts
3geek14: Changed to `ingAsRecipe` (in the new location).

backend/src/services/api-service/optimal/optimal-service.test.ts
```diff
@@ -53,6 +53,36 @@ describe('findOptimalSetsForMeal', () => {
   });
 });
 
+describe('findOptimalMonsForIngredient', () => {
+  it('shall find all optimal mons for an ingredient', () => {
```
SleepAPI, line 57: Yeah I think this works for now, looks like Slowbro and Slowking gather 1.5 tails each here so should be enough margin for now. I guess going forward we would probably want to just mock the call to set-cover class here and see that it is called with the correct parameters, since the functionality of the set-cover itself is outside this function's responsibility. Then assume the set cover handles its own testing (it does).

But yeah this is also how the other tests here work, it's fine for now

backend/src/services/website-converter/website-converter-service.ts
```diff
@@ -160,6 +164,52 @@ class WebsiteConverterServiceImpl {
...
+    const prettifiedCombinations = optimalMons.teams.slice(0, 500).map((solution) => ({
+      team: solution.team
+        .map(
+          (member) =>
+            `${member.pokemonCombination.pokemon.name}(${shortPrettifyIngredientDrop(
+              member.pokemonCombination.ingredientList
+            )})`
+        )
+        .join(', '),
+      details: `👨🏻‍🍳 Team finder - https://sleepapi.net 👨🏻‍🍳\n\nRecipe: ${optimalMons.ingredient} (${
```
SleepAPI, line 177: By default the API returns in proper compact JSON, so this wouldn't be included in the default API anyway unless someone specified pretty output. When you implement the tool on the site, however, this will probably be included in the details section if you're using prettified API. In that case you'll want to update the description here to mention the ingredient tool rather than Team finder. 

Also further down it mentions "requires X mon for this recipe" or something, but this tool always runs with 1 ingredient so the answer would always be 1 mon. Also mentions like "no possible team combinations" which doesn't fit well since we're not finding teams here, but individual mons.

I'm fine keeping this as is for current version though, route isn't used currently and this is only the prettified formatting. Just mentioning it might need another look when the frontend tool is put in place.

SleepAPI (approved):

backend/src/services/api-service/optimal/optimal-service.test.ts
3geek14: After adding Ingredient Magnet mons to the results, I've filtered this list to only Pokemon with at least 1 full ingredient for the test assertions.

backend/src/services/website-converter/website-converter-service.ts
3geek14: I made some more edits to the text now anyway. Hopefully I won't need to make many more changes when I actually implement a page for this.

SleepAPI (approved):

PR #304, opened 2024-05-31
WIP: add events to production calculator

3geek14:
This PR seeks to add events to the production calculator. In Eevee Week, Raikou Research, and Entei Research, we saw buffs to Pokemons' stats, like skill rates. These impact Pokemons' productions, so SleepAPI's production calculator ideally would support this.

We don't know what sort of events Select Button will run in the future, or how they might influence things. This PR handles event buffs with a function passed to the production calculator that modifies a Pokemon. When there's no active event, this function is simply the identity function: `(mon: Pokemon) => mon`. For other events, this function is a little more complicated. The ingredient drop boost in Raikou and Entei was complicated enough to warrant a helper function, which is in `event-util.ts`.

Another approach would be to add more inputs to the production calcualtor to handle events. So far, we've only seen a small number of changes made: increased skill rate, increased ingredient drops, and restricting bonuses to a specific type. I don't like this approach, because I think it risks making it harder to maintain the codebase as new events have new mechanics. If a future event restricts some bonuses to one type and other bonuses to another type, then we'd have to rewrite anything that assumes all bonuses will either go to all Pokemon or to only Pokemon of a specific type.

**WIP Status:**

The first commit adds the `Event` data type, and it adds the events we've seen so far. Future commits will add the ability to pass an event into the production calculator, add an event parameter to the production calculator API route, and add an API route to get a list of all events.

### DIFF
--- a/common/src/domain/event/event.ts
+++ b/common/src/domain/event/event.ts
@@ -1,0 +1,92 @@
+import { AddOneIngredient } from '../../utils';
+import { GREPA, LEPPA } from '../berry';
+import { Pokemon } from '../pokemon';
+
+export interface Event {
+  name: string;
+  description: string;
+  startDate: Date;
+  endDate: Date;
+  statModifier: (mon: Pokemon) => Pokemon;
+}
+
+export const FIRST_ANNIVERSARY_FEST: Event = {
+  name: 'First Anniversary Fest',
+  description: "The chance of each helper Pokemon's main skill triggering will be multiplied by 1.5.",
+  startDate: new Date(2024, 7, 15),
+  endDate: new Date(2024, 7, 21),
+  statModifier: (mon: Pokemon) => {
+    return {
+      skillPercentage: mon.skillPercentage * 1.5,
+      ...mon,
+    };
+  },
+};
+
+export const ENTEI_RESEARCH: Event = {
+  name: 'Entei Research',
+  description:
+    'Fire-type Pokemon find an extra ingredient every time they find ingredients, and their main ' +
+    'skill triggers 50% more often. During the first week, fire-type Pokemon have their main ' +
+    'skill level increased by 1; during the second week, it increases by 3.',
+  startDate: new Date(2024, 5, 20),
+  endDate: new Date(2024, 6, 3),
+  statModifier: (mon: Pokemon) => {
+    if (mon.berry != LEPPA) {
+      return mon;
+    }
+    return {
+      ingredient0: AddOneIngredient(mon.ingredient0),
+      ingredient30: mon.ingredient30.map(AddOneIngredient),
+      ingredient60: mon.ingredient60.map(AddOneIngredient),
+      skillPercentage: mon.skillPercentage * 1.5,
+      ...mon,
+      // For week 1, Main Skill Level is increased by 1
+      // For week 2, Main Skill Level is increased by 3
+    };
+  },
+};
+
+export const RAIKOU_RESEARCH: Event = {
+  name: 'Raikou Research',
+  description:
+    'Electric-type Pokemon find an extra ingredient every time they find ingredients, and their ' +
+    'main skill triggers 50% more often.',
+  startDate: new Date(2024, 3, 25),
+  endDate: new Date(2024, 4, 7),
+  statModifier: (mon: Pokemon) => {
+    if (mon.berry != GREPA) {
+      return mon;
+    }
+    return {
+      ingredient0: AddOneIngredient(mon.ingredient0),
+      ingredient30: mon.ingredient30.map(AddOneIngredient),
+      ingredient60: mon.ingredient60.map(AddOneIngredient),
+      skillPercentage: mon.skillPercentage * 1.5,
+      ...mon,
+    };
+  },
+};
+
+export const EEVEE_WEEK: Event = {
+  name: 'Eevee Week',
+  description: 'Main skills trigger 50% more often.',
+  startDate: new Date(2023, 11, 20),
+  endDate: new Date(2023, 11, 27),
+  statModifier: (mon: Pokemon) => {
+    return {
+      skillPercentage: mon.skillPercentage * 1.5,
+      ...mon,
+    };
+  },
+};
+
+export const NO_EVENT: Event = {
+  name: 'No Event',
+  description: 'No changes.',
+  startDate: new Date(0),
+  endDate: new Date(Infinity),
+  statModifier: (mon: Pokemon) => mon,
+};
+
+export const EVENTS: Event[] = [FIRST_ANNIVERSARY_FEST, ENTEI_RESEARCH, RAIKOU_RESEARCH, EEVEE_WEEK];
--- a/common/src/domain/event/index.ts
+++ b/common/src/domain/event/index.ts
@@ -1,0 +1,1 @@
+export * from './event';
--- a/backend/src/domain/computed/production.ts
+++ b/backend/src/domain/computed/production.ts
@@ -1,4 +1,4 @@
-import { berry, nature, Recipe, subskill } from 'sleepapi-common';
+import { Event, Recipe, berry, nature, subskill } from 'sleepapi-common';
 import { CustomPokemonCombinationWithProduce } from '../combination/custom';
 import { Time } from '../time/time';
 
@@ -22,6 +22,7 @@ export interface ProductionStats {
   mainBedtime: Time;
   mainWakeup: Time;
   maxPotSize?: number;
+  event?: Event;
 }
 
 export interface SetCoverProductionStats extends ProductionStats {
--- a/common/src/domain/index.ts
+++ b/common/src/domain/index.ts
@@ -1,5 +1,6 @@
 export * as berry from './berry';
 export * from './constants';
+export * from './event';
 export * as ingredient from './ingredient';
 export * as island from './island';
 export * as mainskill from './mainskill';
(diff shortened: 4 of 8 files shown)


common/src/domain/event/event.ts
```diff
@@ -0,0 +1,69 @@
+import { GREPA, LEPPA } from 'src/domain/berry';
+import { AddOneIngredient } from 'src/domain/event/event-util';
+import { Pokemon } from 'src/domain/pokemon';
+
+export interface Event {
+  name: string;
+  startDate: string;
```
SleepAPI, line 7: What do you think about Date vs string here? 

Date would enforce a correct format and it would mean we don't have to parse them all to date when listing current/future events in the frontend. 

Using string means we can list the dates in frontend without formatting a Date. 

If we go with Date we should be able to create an event like 
`new Date('2024-01-31')`

common/src/domain/event/event.ts
```diff
@@ -0,0 +1,69 @@
...
+  endDate: string;
+  statModifier: (mon: Pokemon) => Pokemon;
+}
+
+export const ENTEI_RESEARCH: Event = {
```
SleepAPI, line 12: I guess listing all the events in here will eventually become messy. Do you think we could take a similar approach to how we do pokemon? Perhaps one folder for each year, so right now we would have 2023 and 2024

webbywisp (changes requested):

common/src/domain/event/event-util.ts
```diff
@@ -0,0 +1,8 @@
+import { IngredientSet } from 'src/domain';
```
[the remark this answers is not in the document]
3geek14: Done

common/src/domain/event/event.ts
```diff
@@ -0,0 +1,69 @@
+import { GREPA, LEPPA } from 'src/domain/berry';
+import { AddOneIngredient } from 'src/domain/event/event-util';
+import { Pokemon } from 'src/domain/pokemon';
+
+export interface Event {
```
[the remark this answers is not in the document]
3geek14: I've added some descriptions, but I'm not very attached to them. Happy to let other people change these later.

common/src/domain/event/event.ts
3geek14: Thanks, I much prefer the `Date` type.

common/src/domain/event/event.ts
3geek14: I'm not convinced this is necessary, as we've still got way way fewer events than Pokemon. I'd be inclined to punt this until it becomes more of an issue?

SleepAPI: That's fine, this isn't a big issue to me

common/src/domain/event/index.ts
SleepAPI, line 1: Since the domain/event folder is new it needs to be exported from src/domain as well. Currently this index file exports it out of domain/event, but the level above doesnt pick it up

webbywisp (changes requested):
One final comment from my side, regarding the export of domain/event

webbywisp (dismissed):
LGTM, just needs the rebase and merge

3geek14:
I want to push the event adjustment all the way into `setupAndRunProductionSimulation`. Doing that means changing the `PokemonIngredientSet` type to have "AAC" (for example) rather than a full `IngredientSet[]`. That would be a rather large refactor.

If I don't do that, it's easy to get `calculatePokemonProduction` in `production-service.ts` to use the `Event` it receives and pass the modified `Pokemon` into `setupAndRunProductionSimulation`. Similarly, `calculateOptimalProductionForSetCover` has a `ProductionStats` object and can make use of the `Event`.

However, `getAllOptimalIngredientFocusedPokemonProduce` in `contribution-calculator.ts` (used for the tier list) doesn't have a full `ProductionStats` object, so I can't do anything there. This is probably fine, but I think it'd be cool just as a curiosity to see what changes events make to the tier lists.

I think I'll try to do the refactoring. From a "don't repeat yourself" perspective, it feels wrong to deal with event modifications in multiple places. If tier lists don't want to deal with events, that's fine; they can simply not pass in any `Event`, and `setupAndRunProductionSimulation` can apply `NO_EVENT`.

3geek14:
I started working on the refactor, and I ran into an error as soon as I tried to import `Event` into the backend. I've pushed the broken commit in case you want to be able to reproduce the issue. Here's the error message I got from `npm run build`:
```
There was a problem resolving type of 'Event'.
There was a problem resolving type of 'ProductionStats'.
Generate swagger error.
 GenerateMetadataError: Unknown type: FunctionType
At: ~/SleepAPI/common/dist/domain/event/event.d.ts:7:7.
This was caused by 'statModifier: (mon: Pokemon) => Pokemon;'
    at TypeResolver.resolve (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:480:19)
    at TypeResolver.propertyFromSignature (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1195:119)
    at ~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1160:39
    at Array.map (<anonymous>)
    at TypeResolver.getModelProperties (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1160:18)
    at TypeResolver.getModelReference (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1054:33)
    at calcReferenceType (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:920:50)
    at TypeResolver.getReferenceType (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:933:24)
    at TypeResolver.resolve (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:513:36)
    at TypeResolver.propertyFromSignature (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1195:119)
```

SleepAPI:
> I started working on the refactor, and I ran into an error as soon as I tried to import `Event` into the backend. I've pushed the broken commit in case you want to be able to reproduce the issue. Here's the error message I got from `npm run build`:
> 
> ```
> There was a problem resolving type of 'Event'.
> There was a problem resolving type of 'ProductionStats'.
> Generate swagger error.
>  GenerateMetadataError: Unknown type: FunctionType
> At: ~/SleepAPI/common/dist/domain/event/event.d.ts:7:7.
> This was caused by 'statModifier: (mon: Pokemon) => Pokemon;'
>     at TypeResolver.resolve (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:480:19)
>     at TypeResolver.propertyFromSignature (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1195:119)
>     at ~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1160:39
>     at Array.map (<anonymous>)
>     at TypeResolver.getModelProperties (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1160:18)
>     at TypeResolver.getModelReference (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1054:33)
>     at calcReferenceType (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:920:50)
>     at TypeResolver.getReferenceType (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:933:24)
>     at TypeResolver.resolve (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:513:36)
>     at TypeResolver.propertyFromSignature (~/SleepAPI/backend/node_modules/@tsoa/cli/dist/metadataGeneration/typeResolver.js:1195:119)
> ```

This is because ProductionStats interface is returned on the API for some routes. For example both the api/optimal/meal/{name} and your own api/optimal/ingredient/{name} return an attribute called filter of type `SetCoverProductionStats` which just extends ProductionStats. So Swagger is complaining that we can't return a function on the API, since Event type contains statModifier which is a typescript function and not compatible with a HTTP response

3geek14:
Bluh...okay I understand the error now. I guess one option is to have the `ProductionStats` store a `string` that's the name of an `Event`? And then do a lookup. That would prevent power-users from writing their own custom event functions, which is probably desirable, to prevent us from running random user code that we don't trust.

SleepAPI:
> Bluh...okay I understand the error now. I guess one option is to have the `ProductionStats` store a `string` that's the name of an `Event`? And then do a lookup. That would prevent power-users from writing their own custom event functions, which is probably desirable, to prevent us from running random user code that we don't trust.

I think in general for the API incoming request we use ProductionRequest interface. So for incoming requests and lookup we can add an event identifier such as an id or name or something like ENTEI_2024 as id would be cool. 

I think the issue with ProductionStats is that today our APIs are too verbose, we probably shouldn't just return the entire ProductionStats interface out. I think we return some stuff like subskills, nature etc so that the old frontend could use it to list input settings on the results page. We probably should just add event to ProductionStats, but return more explicit type from the APIs. Might need a little refactoring though.

I believe the new frontend doesn't need the input settings returned at all and the team simulator doesn't. The new frontend sends the input settings to the backend and so it already has those beforehand ready.

Issue #357, opened 2024-07-23
Good-Night Ribbons


3geek14:
### Details

A Pokemon can have a Good-Night Ribbon, level 1 to 4,  based on how many hours the player has slept with that Pokemon on the team. The ribbon increases the Pokemon's inventory size. If the Pokemon can evolve, the ribbon can also decrease its frequency.

* `calculator/production` and `calculator/team` APIs should accept ribbon data. The team API should accept different ribbons for different team members.
* Add logic to modify a Pokemon's stats based on its ribbon.
* Add data for which Pokemon can evolve, and how many more times they can evolve.

PR #358, opened 2024-07-23
Add remaining evolutions

3geek14:
This is preparing for #357

### DIFF
--- a/common/src/utils/pokemon-utils/evolution-utils.ts
+++ b/common/src/utils/pokemon-utils/evolution-utils.ts
@@ -1,0 +1,15 @@
+import { Pokemon } from '../../domain/pokemon';
+
+export function evolvesFrom(previousForm: Pokemon): Pokemon {
+  return {
+    ...previousForm,
+    remainingEvolutions: previousForm.remainingEvolutions - 1,
+  };
+}
+
+export function evolvesInto(nextForm: Pokemon): Pokemon {
+  return {
+    ...nextForm,
+    remainingEvolutions: nextForm.remainingEvolutions + 1,
+  };
+}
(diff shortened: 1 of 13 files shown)


common/src/utils/pokemon-utils/evolution-utils.ts
SleepAPI, line 1: Really cool idea! Missing tests for this

SleepAPI (changes requested):
Just missing testing on the new evolutions-utils. Really great stuff. here

common/src/utils/pokemon-utils/evolution-utils.ts
3geek14: I have integration tests in pokemon.test.ts, and I think those are better tests than what we can get from unit testing these functions. Even as these functions grow (from #359, and possibly if we add other stuff later), I think they're simple enough that they might not benefit from unit tests.

I've added two tests. I'm happy to remove them if you don't like them. If there's another type of unit test you'd rather see, let me know?

3geek14: Note that the [code coverage report](https://app.codecov.io/gh/SleepAPI/SleepAPI/pull/358?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=McMomo) from yesterday shows that the only untested line is in index.ts.

SleepAPI: Yes it was tested indirectly, but I personally prefer testing directly too for several reasons, but some bigger reasons are traceability and code standard. By standard we always accompany new functionality by test files. When it comes to traceability if the test in pokemon.test.ts fails you'd have to debug why, in practice it's probably quite easy and this is a smaller issue here, but if we add tests in evolution-utils.test.ts and that specifically fails then it's even easier to pinpoint the exact failing function.

I saw you added tests though and they look good to me!

SleepAPI (approved):

PR #370, opened 2024-08-01
add gen 9 middle stages as optimal

3geek14:
These Pokemon change types when they evolve, so the fully-evolved forms aren't strict upgrades. In particular, some people prefer Crocalor over Skeledirge, as both produce plenty of apples, and Crocalor has an island.

WIP: haven't updated tier lists.


SleepAPI:
I see it's draft, but LGTM so far. Assuming you're trying to push tier list update here as well

PR #395, opened 2024-09-07
add genders to Pokemon

3geek14:
Each Pokemon species has a gender ratio. I looked up each ratio on Bulbapedia. Each Pokemon also has a gender, which should be an enum, so I added a type for that, though it's unused at this time. Future changes to the new frontend will want to use that type for specific Pokemon.

Sometimes, event-spefic Pokemon can have gender ratios different from the species' ratio. Cosplay Pikachu, from ORAS, is always female, despite Pikachu typically being a 50/50 split. It's possible that something similar could be true for Halloween and Holiday Pikachus in Pokemon Sleep. I've added file comments pointing out that those two gender ratios are currently unverified.

Included a test to confirm that, for every Pokemon in the Pokedex, the gender ratios either add up to 0 (for Pokemon with unknown gender) or to 1 (for everything else).


SleepAPI (approved):
Looks great as always

SleepAPI (approved):

PR #411, opened 2024-10-03
fix: coffee longName

extraBottle:
coffee longName fix from "Soothing Cacao" -> "Rousing Coffee"


extraBottle (commented):
added ":" to the request title

PR #506, opened 2025-01-26
Sort pokemon util

jlau3:
Changes Made:
- Updated Rufflet's lvl 30 corn from 1 to 2
- Added pokedexNumbers to all fixtures in berry/ingredient/skill-pokemon.ts
- Added sort-utils along with tests.
-- Made it so you can pass in the field you want to sort by as well as the list of Pokemon whenever filtering comes around (eg. sorting by just berry Pokemon)

- Cleaned up some mocks so that they're more resilient to changes to the Pokemon interface

Considerations:
- Not sure what to do with this bun file, let me know if it should be removed from the PR.
- The sort function currently relies on using the name as a tiebreaker, might need to consider if this needs changing as we have pokemon like witch hat Pikachu that share a name


jlau3:
feel free to rip into this so I can know proper good practices for TS

common/src/domain/pokemon/skill-pokemon.ts
```diff
@@ -720,6 +760,7 @@ export const BONSLY: Pokemon = {
 
 export const DRIFLOON: Pokemon = {
```
3geek14, line 761: Oops, looks like this has been out-of-order for a while! Can you add another commit putting Drifloon and Drifblim between Wynaut and Bonsly?

backend/package.json
3geek14, line 1: I'm surprised this file has any changes. Is it possible that you don't have the same linter settings setup?

common/src/utils/sort-utils/sort-utils.ts
```diff
@@ -0,0 +1,15 @@
+import type { Pokemon } from '../../domain/pokemon';
+
+type sortOrder = 'asc' | 'desc';
+
+export function sortPokemonByField(pokemonList: Pokemon[], field: keyof Pokemon, order: sortOrder = 'asc'): Pokemon[] {
+  return pokemonList.sort((pokemonA, pokemonB) => {
+    if (pokemonA[field] < pokemonB[field]) {
+      return order === 'asc' ? -1 : 1;
+  }
```
3geek14, line 9: It looks like you're missing a level of indentation starting here.

3geek14 (changes requested):
I'll let McMomo comment on the bun file. My suspicion is that it should be reverted. I checked all the Pokedex numbers themselves, and they look right.

SleepAPI:
It seems none of the commits are following conventional commits, you can read about it here:
https://github.com/nerolis-lab/nerolis-lab/blob/develop/CONTRIBUTING.md#commit-standard

We also prefer linear history, so no merge commits, just rebasing your commits on top of latest develop. I won't enforce this as much as conventional commits though, just a preference. I know @tindomerel8 also likes merge commits.

backend/package.json
SleepAPI: That said there's a good chance this PR will fail linting when we run the pipeline. So @jlau3 might either need to save the package.json to reformat it, or add it to exclusion rules as mentioned in my comment above this

backend/bun.lockb
SleepAPI, line 1: That's strange, this should not be needed. This is the old way of bun creating lock files.
We use the new way, you should see a bun.lock file at the same location.

`bun install` should grab the pre-existing bun.lock file and not create a new bun.lockb. If that doesn't work then try `bun install --save-text-lockfile`. 

Anyway, this lockb should not be needed and we shouldn't check it in.

common/src/utils/carry-size-utils/carry-size-utils.test.ts
```diff
@@ -220,22 +220,12 @@ describe('calculateCarrySize', () => {
...
 describe('base and max carry size', () => {
-  const MOCK_POKEMON: Pokemon = {
-    name: 'Mockemon',
-    specialty: 'berry',
-    frequency: 0,
-    ingredientPercentage: 0,
-    skillPercentage: 0,
-    berry: BELUE,
-    genders: BALANCED_GENDER,
+  const MOCK_POKEMON: Pokemon = mockPokemon({
```
SleepAPI, line 223: Love this, good cleanup, thanks!

common/src/utils/sort-utils/sort-utils.ts
SleepAPI: Yeah this file does not seem to be formatted. This leads me to believe @jlau3 does not have the autoformatting set up.

This is explained in step 2 and 3 in the [DEVELOPMENT_SETUP.md](https://github.com/nerolis-lab/nerolis-lab/blob/develop/DEVELOPMENT_SETUP.md#2-open-workspace-in-vscode). Essentially you need to npm install in the root of the project and open the workspace file in vs-code.

common/src/utils/sort-utils/sort-utils.test.ts
SleepAPI, line 1: Love this, excellent tests.

common/src/utils/sort-utils/sort-utils.ts
SleepAPI, line 1: Love this idea with the sort-utils btw, this is a very cool and reusable util.

SleepAPI (changes requested):
Couple minor requests, but mostly this is insanely good. 
Great work!

jlau3:
Made a separate PR cleaning up commit messages and working on the develop branch
https://github.com/nerolis-lab/nerolis-lab/pull/507

PR #507, opened 2025-01-27
Added pokedexNumbers, cleaned up some mocks, and added sorting by field

jlau3:
Changes Made:

Updated Rufflet's lvl 30 corn from 1 to 2

Added pokedexNumbers to all fixtures in berry/ingredient/skill-pokemon.ts

Added sort-utils along with tests.
-- Made it so you can pass in the field you want to sort by as well as the list of Pokemon whenever filtering comes around (eg. sorting by just berry Pokemon)

Cleaned up some mocks so that they're more resilient to changes to the Pokemon interface

Considerations:
The sort function currently relies on using the name as a tiebreaker, might need to consider if this needs changing as we have pokemon like 2 versions of witch hat Pikachu that share a name


common/src/utils/sort-utils/sort-utils.ts
```diff
@@ -0,0 +1,15 @@
+import type { Pokemon } from '../../domain/pokemon';
+
+type sortOrder = 'asc' | 'desc';
+
+export function sortPokemonByField(pokemonList: Pokemon[], field: keyof Pokemon, order: sortOrder = 'asc'): Pokemon[] {
```
SleepAPI, line 5: minor comment: no need to fix this in this PR, since it's really no big deal, but this would create slightly stronger typing for keyof Pokemon:
```typescript
export function sortPokemonByField<T extends keyof Pokemon>(pokemonList: Pokemon[], field: T, order: sortOrder = 'asc'): Pokemon[] {
```

SleepAPI (changes requested):
Seems like backend tests are failing. Backend and frontend both use common, so changes a central interface such as Pokemon is likely to have affects on backend/frontend. 

You should be able to `npm i && npm run build` in both of those and find the compilation errors, but please let me know if you need help. After that just `npm run test` just like you did in common, to find the breaking tests.

SleepAPI (approved):

PR #512, opened 2025-01-28
feat: Created displayName for all pokemon

jlau3:
- Added displayName to all Pokemon.
-- Used the name under the sleep style dex for Pokemon with alternate forms. eg. Pikachu (Holiday) and the like
- Fixed some mocks to more accurately display what the name field should look like 'Mockemon' -> 'MOCKEMON' (no real feature change here, just being particular)
- Changed asset files to reflect the proper name of the Pokemon

Considerations:
- Frontend needs to be updated to use the displayNames wherever necessary


SleepAPI (changes requested):
Good idea with renaming the name and updating the assets accordingly. I believe you're missing the backend assets though. They are used in the alpha site `sleepapi.net` as long as it's still up. 

They're over at backend/src/assets/pokemon.

You might want to run the backend and frontend locally. You'll be able to reach the alpha site at localhost:3000 and Neroli's Lab at localhost:8001. That should let you check if the assets work

SleepAPI (approved):

PR #543, opened 2025-02-12
Dynamic calculations for ingredient max recipe bonuses

jlau3:
Changes:
- Wrapped Ingredients in a createIngredient function
- Created an ingredientBonusCache that's updated when recipes are created
- Updated frontend to use the ingredientBonusCache instead of MAX_RECIPE_BONUS constant
- Removed MAX_RECIPE_CONSTANT

Considerations:
- Right now, within frontend, if ingredientBonusCache can't find an ingredient, it will default to 0 silently, do we want to update this interaction to be logged?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [1.31.0](https://github.com/nerolis-lab/nerolis-lab/compare/v1.30.2...v1.31.0) (2025-02-10)
+
+
+### Features
+
+* adds clodsire and paldean wooper ([88e13f5](https://github.com/nerolis-lab/nerolis-lab/commit/88e13f55e4d6547afdea5e0470f3bf5e39885a38))
+
 ## [1.30.2](https://github.com/nerolis-lab/nerolis-lab/compare/v1.30.1...v1.30.2) (2025-02-09)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "sleepapi",
-  "version": "1.30.2",
+  "version": "1.31.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "sleepapi",
-      "version": "1.30.2",
+      "version": "1.31.0",
       "devDependencies": {
         "@eslint/js": "^9.16.0",
         "@eslint/json": "^0.8.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sleepapi",
-  "version": "1.30.2",
+  "version": "1.31.0",
   "private": true,
   "type": "module",
   "scripts": {


common/src/domain/ingredient/ingredients.ts
```diff
@@ -138,4 +53,14 @@ export const INGREDIENTS: Ingredient[] = [
   LARGE_LEEK,
   SLOWPOKE_TAIL
 ];
+
 export const TOTAL_NUMBER_OF_INGREDIENTS = INGREDIENTS.length;
+
+export function createIngredient(name: string, value: number, taxedValue: number, longName: string): Ingredient {
```
SleepAPI, line 59: Can we move this to `ingredient-utils.ts` and the test to `ingredient-utils.test.ts`?
I realize `ingredient-utils.ts` is somewhat big, could use some cleanup eventually, but let's not do that now.

Can we also params-wrap the parameters?
```typescript
export function createIngredient(params: { name: string, value: number, taxedValue: number, longName: string }): Ingredient {
const { name, value, taxedValue, longName } = params;
```
That allows us to create ingredients like:
```typescript
export const FANCY_EGG: Ingredient = createIngredient({name: 'Egg', value: 115, taxedValue: 38.7, longName: 'Fancy Egg'});
```
The main reason is that we are able to tell which parameters we're actually sending in:
```typescript
export const HONEY: Ingredient = createIngredient('Honey', 101, 29.8, 'Honey');
```
Here it's not completely obvious what each parameter represents, there's two Honey and two numbers. The params would fix that, especially useful for new devs

common/src/utils/recipe-utils/recipe-utils.test.ts
```diff
@@ -199,3 +206,50 @@ describe('calculateRecipeValue', () => {
     expect(valueLevel10).toBeGreaterThan(valueLevel1);
   });
 });
+
+const mockRecipeList = [
```
SleepAPI, line 210: We have mock factory helpers for creating recipes, so you can do:
```typescript
const mockRecipeList = [
commonMocks.recipe({ name: 'MOCK_RECIPE_WEAK_BONUS_UNUPDATED',
    ingredients: [
      { amount: 10, ingredient: SLOWPOKE_TAIL },
      { amount: 20, ingredient: SOFT_POTATO }
    ],
    bonus: 17.55 }),
    commonMocks.recipe({{
    name: 'MOCK_RECIPE_WEAK_BONUS',
    ingredients: [{ amount: 4, ingredient: PURE_OIL }],
    bonus: 15.77
  }})
  ... // add rest
  ```

File: common/src/utils/recipe-utils/recipe-utils.ts [hunk 3]
```diff
@@ -4,6 +4,8 @@ import type { Recipe, RecipeFlat, RecipeType } from '../../domain/recipe';
 import { emptyIngredientInventoryFloat } from '../../utils/flat-utils';
 import { ING_ID_LOOKUP } from '../ingredient-utils/ingredient-utils';
 
+export const ingredientBonusCache = new Map<string, number>();
```
SleepAPI, line 7: I'm not a big fan of exposing the cache directly, can we move both the cache and `updateIngredientBonus` to `ingredient-utils.ts` and remove the export from the cache?

Then we should add a new util, something like `getMaxIngredientBonus()` that performs the lookup in cache, and now we can also log error if an ingredient lookup fails.

frontend/src/components/compare/compare-strength.vue
```diff
@@ -356,16 +356,14 @@ export default defineComponent({
...
       const amount =
-        recipeBonus *
         maxLevelRecipeMultiplier *
         this.userStore.islandBonus *
-        memberProduction.produceTotal.ingredients.reduce(
-          (sum, cur) => sum + cur.amount * cur.ingredient.value * AVERAGE_WEEKLY_CRIT_MULTIPLIER,
-          0
-        )
+        memberProduction.produceTotal.ingredients.reduce((sum, cur) => {
+          const ingredientBonus = 1 + (ingredientBonusCache.get(cur.ingredient.name) ?? 0) / 100
```
SleepAPI, line 364: This is correct! Just remember to update this to use the util instead of cache directly, that would also mean you no longer have to default zero here, you can do that in the util.

SleepAPI (changes requested):
Honestly just some minor code style comments, the functionality is there!
Great job on this 💪

common/src/domain/ingredient/ingredients.ts
jlau3: I've implemented these changes but I'd like to keep createIngredient within `ingredient.ts`. There's a circular dependency between `ingredient.ts` and `ingredient-utils.ts` because of the usage of the `INGREDIENTS` and `TOTAL_NUMBER_OF_INGREDIENTS` list and constant in the utils file.

I can move createIngredient to `ingredient-utils.ts` if we move those 2 consts out as well. Open to other ideas

SleepAPI: Right, that's fair enough, let's keep it in `ingredients.ts` for now then 👍

jlau3:
Hmm....Prettier checks didn't fail locally when I built and tested this. Is there a command I'm missing for linting the project files?

SleepAPI:
> Hmm....Prettier checks didn't fail locally when I built and tested this. Is there a command I'm missing for linting the project files?

Yes you can run `npm run lint` in the root of the project. You should never have to do so though, if the lint check is failing then that means you don't have the autoformatting set up.

Are you using VS code? It is recommended to use VS Code and open the workspace file in the root of the project.
After that you run `npm i` in the root of the project and install the prettier and eslint extension. Now everything should autoformat as soon as you save a file, which means you will always pass linting.

Here is the section in the [DEVELOPMENT_SETUP](<https://github.com/nerolis-lab/nerolis-lab/blob/develop/DEVELOPMENT_SETUP.md#2-open-workspace-in-vscode>). When you open the workspace you *should* be prompted to automatically download the recommended extensions, but if not, or if you have already opened the workspace before, you can find them in the bottom of the workspace file.

common/src/utils/ingredient-utils/ingredient-utils.ts
```diff
@@ -214,3 +216,17 @@ export function calculateAveragePokemonIngredientSet(
...
+    const ingredientName = ingredientSet.ingredient.name;
+    const currentBonus = ingredientBonusCache.get(ingredientName);
+    if (currentBonus === undefined || currentBonus < bonus) {
+      ingredientBonusCache.set(ingredientName, bonus);
+    }
+  }
+}
+
+export function getMaxIngredientBonus(ingredientName: string): number {
+  return ingredientBonusCache.get(ingredientName) ?? 0;
```
SleepAPI, line 231: Perhaps we should log an error if the get returns undefined here

jlau3:
Yea, I do have the workspace open with all the extensions installed so I'm sort of confused how this happened myself

SleepAPI (changes requested):
Could we squash the prettier fix commits? From the develop branch perspective those commits fixed issues created within the same pull request, thus they are kind of irrelevant to the develop branch imo

jlau3:
ugh, I messed up my rebase, will fix

PR #545, opened 2025-02-13
feat: dynamic max ingredient bonuses instead of static ones

jlau3:
Changes:

    Wrapped Ingredients in a createIngredient function
    Created an ingredientBonusCache that's updated when recipes are created
    Updated frontend to use the ingredientBonusCache instead of MAX_RECIPE_BONUS constant
    Removed MAX_RECIPE_CONSTANT

### DIFF
--- a/common/src/utils/ingredient-utils/ingredient-utils.ts
+++ b/common/src/utils/ingredient-utils/ingredient-utils.ts
@@ -7,6 +7,7 @@ import type {
 } from '../../domain/ingredient/ingredient';
 import { INGREDIENTS, TOTAL_NUMBER_OF_INGREDIENTS } from '../../domain/ingredient/ingredients';
 import type { Pokemon } from '../../domain/pokemon/pokemon';
+import '../../prototype/logger';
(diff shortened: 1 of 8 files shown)


common/src/utils/ingredient-utils/ingredient-utils.ts
```diff
@@ -7,14 +7,19 @@ import type {
 } from '../../domain/ingredient/ingredient';
 import { INGREDIENTS, TOTAL_NUMBER_OF_INGREDIENTS } from '../../domain/ingredient/ingredients';
 import type { Pokemon } from '../../domain/pokemon/pokemon';
+import { Logger } from '../../prototype';
 import { emptyIngredientInventoryFloat, emptyIngredientInventoryInt } from '../flat-utils';
 import { MathUtils } from '../math-utils/math-utils';
 import { capitalize } from '../string-utils/string-utils';
 
+const logger = new Logger();
```
SleepAPI, line 15: `logger` is declared global in `logger.ts` which means all that is needed is for you to import that file like:
```typescript
import '../../prototype/logger'
```
and then you can use logger, no need to create a new instance 👍

SleepAPI (approved):

PR #548, opened 2025-02-20
feat: Switch member simulator to roll berries/ingredients individually

CarelessParsley:
There is also a crappy UI hookup for the resulting data but it needs more work. The server side itself though I think is ready.

### DIFF
--- a/backend/src/services/simulation-service/team-simulator/cooking-state/cooking-state.ts
+++ b/backend/src/services/simulation-service/team-simulator/cooking-state/cooking-state.ts
@@ -1,3 +1,5 @@
+import { createPreGeneratedRandom } from '@src/utils/random-utils/pre-generated-random.js';
+import type { PreGeneratedRandom } from '@src/utils/random-utils/pre-generated-random.js';
 import type {
   UserRecipeFlat,
   UserRecipes
@@ -5,7 +7,6 @@ import type {
 import type { CookedRecipeResult, CookingResult, IngredientIndexToFloatAmount, TeamSettingsExt } from 'sleepapi-common';
 import {
   MAX_POT_SIZE,
-  RandomUtils,
   curry,
   dessert,
   emptyIngredientInventoryFloat,
@@ -34,6 +35,7 @@ export class CookingState {
   private bonusCritChance = 0;
   private totalCritChance = 0;
   private totalWeekdayPotSize = 0;
+  private rng: PreGeneratedRandom;
 
   private userCurries: UserRecipeFlat[];
   private userSalads: UserRecipeFlat[];
@@ -56,11 +58,12 @@ export class CookingState {
   private currentSaladStockpile: IngredientIndexToFloatAmount;
   private currentDessertStockpile: IngredientIndexToFloatAmount;
 
-  constructor(settings: TeamSettingsExt, userRecipes: UserRecipes) {
+  constructor(settings: TeamSettingsExt, userRecipes: UserRecipes, rng: PreGeneratedRandom | null = null) {
     const { curries, salads, desserts } = userRecipes;
     this.userCurries = curries;
     this.userSalads = salads;
     this.userDesserts = desserts;
+    this.rng = rng || createPreGeneratedRandom();
 
     this.camp = settings.camp;
     this.startingStockpiledIngredients = settings.stockpiledIngredients;
@@ -115,7 +118,7 @@ export class CookingState {
       currentStockpile: this.currentDessertStockpile
     }) ?? { ...dessert.MIXED_JUICE_FLAT, level: 1 };
 
-    const extraTasty = RandomUtils.roll(currentCritChance);
+    const extraTasty = this.rng() < currentCritChance;
     const extraTastyFactor = extraTasty ? (sunday ? 3 : 2) : 1;
 
     if (extraTasty) {
--- a/backend/src/services/simulation-service/team-simulator/member-state/member-state-utils.ts
+++ b/backend/src/services/simulation-service/team-simulator/member-state/member-state-utils.ts
@@ -1,4 +1,6 @@
-export function calculateSkillProcDistribution(data: number[]): Record<number, number> {
+type NumericCollection = ArrayLike<number> & Iterable<number>;
+
+export function calculateDistribution(data: NumericCollection): Record<number, number> {
   if (data.length === 0) {
     return {};
   }
--- a/backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
+++ b/backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
@@ -1,6 +1,7 @@
 import type { SleepInfo } from '@src/domain/sleep/sleep-info.js';
 import { calculateSleepEnergyRecovery } from '@src/services/calculator/energy/energy-calculator.js';
 import type { CookingState } from '@src/services/simulation-service/team-simulator/cooking-state/cooking-state.js';
+import { calculateDistribution } from '@src/services/simulation-service/team-simulator/member-state/member-state-utils.js';
 import type {
   SkillActivationValue,
   TeamSkillActivation
@@ -9,14 +10,18 @@ import { SkillState } from '@src/services/simulation-service/team-simulator/skil
 import { TeamSimulatorUtils } from '@src/services/simulation-service/team-simulator/team-simulator-utils.js';
 import { getMealRecoveryAmount } from '@src/utils/meal-utils/meal-utils.js';
 import { TimeUtils } from '@src/utils/time-utils/time-utils.js';
+import type { PreGeneratedRandom } from '@src/utils/random-utils/pre-generated-random.js';
+import { createPreGeneratedRandom } from '@src/utils/random-utils/pre-generated-random.js';
 import type {
   BerryIndexToFloatAmount,
   BerrySet,
+  IngredientIndexToFloatAmount,
+  IngredientIndexToIntAmount,
+  IngredientSet,
   MemberProduction,
   MemberProductionBase,
   PokemonWithIngredientsIndexed,
   Produce,
-  ProduceFlat,
   SimpleTeamResult,
   TeamMemberExt,
   TeamSettingsExt,
@@ -25,19 +30,19 @@ import type {
 import {
   AVERAGE_WEEKLY_CRIT_MULTIPLIER,
   CarrySizeUtils,
+  ING_ID_LOOKUP,
   MAX_POT_SIZE,
   MathUtils,
   berrySetToFlat,
   calculateNrOfBerriesPerDrop,
   emptyBerryInventoryFloat,
-  flatToBerrySet,
+  emptyIngredientInventoryFloat,
+  emptyIngredientInventoryInt,
   flatToIngredientSet,
-  getEmptyInventoryFloat,
   ingredientSetToFloatFlat,
   ingredientSetToIntFlat,
   multiplyProduce,
-  subskill,
-  sumFlats
+  subskill
 } from 'sleepapi-common';
 
 export class MemberState {
@@ -47,6 +52,7 @@ export class MemberState {
   cookingState?: CookingState;
   private skillState: SkillState;
   private camp: boolean;
+  private rng: PreGeneratedRandom;
 
   // quick lookups, static data
   private pokemonWithIngredients: PokemonWithIngredientsIndexed;
@@ -63,7 +69,19 @@ export class MemberState {
   private helpsSinceLastCook = 0;
   private totalAverageHelps = 0;
   private totalSneakySnackHelps = 0;
-  private voidHelps = 0;
+  private totalBerryProduction = 0;
+  private totalIngredientProduction: IngredientIndexToFloatAmount = emptyIngredientInventoryFloat();
+  private voidIngredients: IngredientIndexToFloatAmount = emptyIngredientInventoryFloat();
+
+  private berryProductionPerDay: Int16Array;
+  private ingredientProductionPerDay: IngredientIndexToIntAmount[];
+  private currentDay = 0;
+  private totalDays: number;
+
+  // ingredient sets
+  private level0IngredientSet: IngredientSet;
+  private level30IngredientSet?: IngredientSet;
+  private level60IngredientSet?: IngredientSet;
 
   // stats
   private frequency0;
@@ -73,12 +91,15 @@ export class MemberState {
   private frequency80;
   public skillPercentage: number;
   private ingredientPercentage: number;
-  private sneakySnackBerries: BerryIndexToFloatAmount = emptyBerryInventoryFloat();
-  private averageProduceAmount: number;
-  private averageProduce: ProduceFlat = getEmptyInventoryFloat();
+
+  // precomputed berry drop amounts
+  private berryDropAmounts: BerryIndexToFloatAmount = emptyBerryInventoryFloat();
+  private berryDropAmount = 0;
 
   // summary
-  totalProduce: Produce = CarrySizeUtils.getEmptyInventory();
+
+  // TODO: move to skill-state
+  skillProduce: Produce = CarrySizeUtils.getEmptyInventory();
   totalRecovery = 0;
   wastedEnergy = 0;
   private energyIntervalsDay = 0;
@@ -100,16 +121,55 @@ export class MemberState {
   private totalHealedByMembers = 0;
   private totalHelpsByMembers = 0;
 
+  // This tracks the actual amount of ingredients accumulated since last cook
+  private ingredientsSinceLastCook = emptyIngredientInventoryFloat();
+
+  // We roll a random number between 0 and 1, then:
+  //
+  //     berry   ing0   ing30   ing60
+  //  |--------|------|-------|--------| 1
+  //  |        |      |       | ingredient60Threshold
+  //  |        |      | ingredient30Threshold
+  //  |        | ingredient0Threshold
+  //  | 0
+  //
+  private ingredient0Threshold: number;
+  private ingredient30Threshold: number;
+  private ingredient60Threshold: number;
+  // Integer thresholds for uint8 comparisons (0-255)
+  private ingredient0ThresholdInt: number;
+  private ingredient30ThresholdInt: number;
+  private ingredient60ThresholdInt: number;
+
+  // precomputed ingredient ids and amounts
+  private level0IngredientId: number;
+  private level30IngredientId?: number;
+  private level60IngredientId?: number;
+  private level0IngredientAmount: number;
+  private level30IngredientAmount?: number;
+  private level60IngredientAmount?: number;
+
   constructor(params: {
     member: TeamMemberExt;
     team: TeamMemberExt[];
     settings: TeamSettingsExt;
     cookingState: CookingState | undefined;
+    iterations?: number;
+    rng?: PreGeneratedRandom;
   }) {
-    const { member, team, settings, cookingState } = params;
+    const { member, team, settings, cookingState, iterations = 1, rng } = params;
+
+    this.rng = rng || createPreGeneratedRandom();
+
+    // Initialize the daily production arrays with the total number of days we'll simulate
+    // Each iteration is one day
+    this.totalDays = iterations;
+    this.berryProductionPerDay = new Int16Array(iterations);
+    this.ingredientProductionPerDay = Array(iterations)
+      .fill(null)
+      .map(() => emptyIngredientInventoryInt());
 
     this.camp = settings.camp;
-
     this.cookingState = cookingState;
 
     // list already filtered for level
@@ -146,6 +206,24 @@ export class MemberState {
       pokemon: member.pokemonWithIngredients.pokemon.name,
       ingredients: pokemonIngredientListFlat
     };
+
+    // Initialize ingredient sets from the Pokemon's actual ingredient list
+    const ingredientList = member.pokemonWithIngredients.ingredientList;
+
+    // Level 0 ingredient is always the first one
+    this.level0IngredientSet = ingredientList[0];
+
+    // Level 30 ingredient is second if available
+    if (member.settings.level >= 30 && ingredientList.length > 1) {
+      this.level30IngredientSet = ingredientList[1];
+    }
+
+    // Level 60 ingredient is third if available
+    if (member.settings.level >= 60 && ingredientList.length > 2) {
+      this.level60IngredientSet = ingredientList[2];
+    }
+
+    // Calculate raw production amounts without probability adjustments
     const memberBerryInList = berrySetToFlat([
       {
         amount: 1,
@@ -157,10 +235,9 @@ export class MemberState {
       member.pokemonWithIngredients.pokemon.specialty,
       member.settings.subskills
     );
-    this.averageProduce = TeamSimulatorUtils.calculateAverageProduce(member);
-    this.averageProduceAmount = sumFlats(this.averageProduce.ingredients, this.averageProduce.berries);
 
-    this.sneakySnackBerries = memberBerryInList.map((amount) => (amount *= berriesPerDrop));
+    this.berryDropAmounts = memberBerryInList.map((amount: number) => amount * berriesPerDrop);
+    this.berryDropAmount = Math.max(...this.berryDropAmounts);
 
     const frequency = TeamSimulatorUtils.calculateHelpSpeedBeforeEnergy({
       member,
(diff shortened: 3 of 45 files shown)


backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -63,8 +68,22 @@ export class MemberState {
...
+  private berryProductionPerDay: number[] = [];
+  private ingredientProductionPerDay: number[] = [];
+  private currentDayBerryProduction = 0;
+  private currentDayIngredientProduction = 0;
+
+  // Ingredient tracking by level
+  private level0IngredientSet: IngredientSet;
+  private level30IngredientSet: IngredientSet | null;
+  private level60IngredientSet: IngredientSet | null;
+  private possibleIngredientLevels: number[] = [];
```
CarelessParsley, line 85: The exact data rep here is easy for me to change, feel free to nitpick if you have preferences.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -74,8 +93,8 @@ export class MemberState {
   public skillPercentage: number;
   private ingredientPercentage: number;
   private sneakySnackBerries: BerryIndexToFloatAmount = emptyBerryInventoryFloat();
-  private averageProduceAmount: number;
-  private averageProduce: ProduceFlat = getEmptyInventoryFloat();
+  private rawProduce: ProduceFlat = getEmptyInventoryFloat();
```
CarelessParsley, line 96: Something funny going on here

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -63,8 +66,30 @@ export class MemberState {
...
+  // got over the Monte Carlo simulations.  The way that MemberState is
+  // interacted with over the simulations is that we repeatedly run the sim
+  // with the *same* MemberState.  So we cannot have a single counter for
+  // number of helps.  Instead, we need a per-day counter, which we then shove
+  // into an array so we know exactly how much we produced each day.
+  private berryProductionPerDay: number[] = [];
+  // Each element is a Float32Array representing all ingredients for that day
+  private ingredientProductionPerDay: Float32Array[] = [];
+  private currentDayBerryProduction = 0;
+  private currentDayIngredientProduction = emptyIngredientInventoryFloat();
```
SleepAPI, line 83: This would then become `emptyIngredientInventoryInt`. The `addIngredients` function should be able to replace float array with int array too since as far as I know we should never be adding decimal value ingredients anymore.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -74,10 +99,18 @@ export class MemberState {
...
+
+  // Pre-compute how many berries the Pokemon produces per drop.  Sometimes
+  // it's convenient to have the full one-hot array, other times its
+  // convenient to just the count.
+  private berryDropAmounts: Float32Array = new Float32Array();
+  private berryDropAmount: number = 0;
 
   // summary
+
+  // NB: Although this /looks/ like the total production over all Monte Carlo
```
SleepAPI, line 111: You're right, this is an artifact, it used to be the total production before we made a significant refactor of the simulator by tracking amount numbers instead to favor performance. 

Let's remove the comment and rename the variable to something like `skillProduce`, also it should probably be moved to `skill-state` and returned from its results, but if that feels out of scope could just rename and leave a `// TODO: move to skill-state` comment for now

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -100,6 +133,9 @@ export class MemberState {
   private totalHealedByMembers = 0;
   private totalHelpsByMembers = 0;
 
+  // This tracks the actual amount of ingredients accumulated since last cook
```
SleepAPI, line 136: Let's rename the variable to indicate ingredients rather than ingredient helps then: `ingredientsSinceLastCook`

SleepAPI: Also—this will become `emptyIngredientInventoryInt`

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -146,6 +181,46 @@ export class MemberState {
...
+    } else {
+      this.level60IngredientSet = null;
+    }
+
+    // Initialize tracking for all ingredients
+    for (const ing of allIngredients) {
+      const ingName = ing.ingredient.name;
+      const ingredientId = ING_ID_LOOKUP[ingName];
+      // Initialize the current day tracking arrays
+      this.currentDayIngredientProduction[ingredientId] = 0;
```
SleepAPI, line 219: Flat arrays already initialize with 0 as default values, so this block is not necessary which then also means `allIngredients` array is not needed anymore.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -146,6 +181,46 @@ export class MemberState {
       pokemon: member.pokemonWithIngredients.pokemon.name,
       ingredients: pokemonIngredientListFlat
     };
+
+    // Pre-calculate ingredient amounts by level using IngredientSet and initialize tracking
+    const allIngredients = [];
+
+    // Initialize ingredient sets from the Pokemon's actual ingredient list
```
SleepAPI, line 188: As per comments above we should be able to replace possibleIngredientLevels array with just a number 0, 1 or 2 which we'll use to random roll. That would probably remove most of this code too.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -288,6 +375,35 @@ export class MemberState {
       this.totalAverageHelps += 1;
       this.helpsSinceLastCook += 1;
 
+      // Roll for berry vs ingredient
+      if (RandomUtils.roll(1 - this.ingredientPercentage)) {
```
SleepAPI, line 379: Okay hear me out, this is a little iffy, but I guess it could optimize somewhat. We could probably reduce the number of random rolls from 2 to 1. Currently this rolls if berry help or ingredient help.

Instead we could probably roll a random number between 1 and 100, if the number is `1 - this.ingredientPercentage` it's a berry help, then we divide the remaining ingredientPercentage into 3 equal chunks and see which ingredient we hit.

Like say we roll 87 and the ingredient percentage was 30%, that means we rolled an ingredient help since berry help would be < 70%. Furthermore it would mean we got the level 30 ingredient since level 0 is 70-80, level 30 is 80-90 and level 60 is 90-100.

Just a stupid random thought, I'm not sure if this would even improve performance actually and if it doesn't then definitely don't do it. I'm not entirely sure how expensive random rolls are.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -157,8 +232,9 @@ export class MemberState {
       member.pokemonWithIngredients.pokemon.specialty,
       member.settings.subskills
     );
-    this.averageProduce = TeamSimulatorUtils.calculateAverageProduce(member);
-    this.averageProduceAmount = sumFlats(this.averageProduce.ingredients, this.averageProduce.berries);
+
+    this.berryDropAmounts = Float32Array.from(memberBerryInList, (value) => value * berriesPerDrop);
```
SleepAPI, line 236: This is the exact same thing as sneakySnackBerries, we can just rename sneakySnackBerries to berryDropAmount and remove this (I prefer the map version)

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -270,9 +355,11 @@ export class MemberState {
...
 
   public updateIngredientBag() {
-    const ingsSinceLastCook = this.averageProduce.ingredients._mapUnary((a) => a * this.helpsSinceLastCook);
-    this.cookingState?.addIngredients(ingsSinceLastCook);
-    this.helpsSinceLastCook = 0;
+    // Since we're tracking actual ingredient amounts in Float32Array, we can directly use it
+    this.cookingState?.addIngredients(this.ingredientHelpsSinceLastCook);
+
+    // Reset ingredient amounts
+    this.ingredientHelpsSinceLastCook = emptyIngredientInventoryFloat();
```
SleepAPI, line 362: I checked this function out again and since we're now calling it MUCH more often it should probably be changed to actually be performant, we should be able to do just:
```typescript
export function initFloatArray(length: number) {
  return new Float32Array(length);
}
```

This goes for initIntArray too of course.

Furthermore the `emptyIngredientInventoryFloat` method (and int and berry inventory methods too). use `INGREDIENTS.length` which we can replace with just `TOTAL_NUMBER_OF_INGREDIENTS` which is a precompiled constant so we can skip the .length check on every call.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -288,6 +375,35 @@ export class MemberState {
...
+      // Roll for berry vs ingredient
+      if (RandomUtils.roll(1 - this.ingredientPercentage)) {
+        // Berry drop
+        this.totalBerryHelps += 1;
+        this.currentDayBerryProduction += 1;
+      } else {
+        // Ingredient drop
+        // Roll for which ingredient level (equal probability)
+        const roll = RandomUtils.randomElement(this.possibleIngredientLevels) ?? 0;
+        let ingredientSet: IngredientSet | null = null;
```
SleepAPI, line 387: This functionality is probably taking quite a bit of the performance. We should be able to change the randomElement as per above comments to just `randomNumber(this.possibleIngredientLevels` which is now just a number instead of an array. This would return a number either 0, 1 or 2. We could pre-calculate an array of the same size which has the amount of each ingredient produced in it.

Then we can just do:
```typescript
const rolledIngredient = RandomUtils.randomNumber(this.possibleIngredientLevels)
const ingredientAmount = this.ingredientHelpAmounts[rolledIngredient]

this.totalIngredientHelps[ingredientId] += ingredientAmount;
this.currentDayIngredientProduction[ingredientId] += ingredientAmount;
this.ingredientHelpsSinceLastCook[ingredientId] += ingredientAmount;
```

SleepAPI: I won't comment on the night helps, same applies there of course

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -288,6 +375,35 @@ export class MemberState {
...
+          // Level 30
+          ingredientSet = this.level30IngredientSet;
+        } else {
+          // Level 0
+          ingredientSet = this.level0IngredientSet;
+        }
+
+        if (ingredientSet) {
+          const ingredientId = ING_ID_LOOKUP[ingredientSet.ingredient.name];
+          this.totalIngredientHelps[ingredientId] += ingredientSet.amount;
```
SleepAPI, line 401: Let's rename these variables since they're not helps, they are actual amounts.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -307,20 +423,58 @@ export class MemberState {
...
+        this.carriedAmount += totalDropAmount;
+
+        // Calculate how much of the help fits in the bag
+        const helpRatio = inventorySpace >= totalDropAmount ? 1 : inventorySpace / totalDropAmount;
+        this.totalAverageHelps += helpRatio;
+        this.helpsSinceLastCook += helpRatio;
+
+        if (isBerryDrop) {
+          // Berry drop
+          this.totalBerryHelps += helpRatio;
```
SleepAPI, line 461: You already branched on if berry help or not above, this seems like unnecessary computing?

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -288,6 +375,35 @@ export class MemberState {
       this.totalAverageHelps += 1;
       this.helpsSinceLastCook += 1;
 
+      // Roll for berry vs ingredient
+      if (RandomUtils.roll(1 - this.ingredientPercentage)) {
+        // Berry drop
+        this.totalBerryHelps += 1;
+        this.currentDayBerryProduction += 1;
```
SleepAPI, line 382: Is this helps or berry amount? You're incrementing with 1 here which indicates helps, but the variable name indicates you're intending to increment by the actual number of produced berries.

Furthermore for `currentDayIngredientAmount` you're incrementing by the actual amount rather than 1. Is the berry histogram supposed to be berry helps and the ingredient histogram supposed to be amounts? Is that disconnect intentional?

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -307,20 +423,58 @@ export class MemberState {
...
+        }
+
+        if (helpRatio < 1) {
+          this.voidHelps += 1 - helpRatio;
         }
       } else {
         this.nightHelpsAfterSS += 1;
         this.totalSneakySnackHelps += 1;
+        // Sneaky snack always gives berries
+        this.totalBerryHelps += 1;
```
SleepAPI, line 477: Hmmm, should sneaky snacking still count as berry helps for the histogram? Should we increment `currentDayBerry` here?

SleepAPI (changes requested):
This is a big change, wow I'm very impressed. There's still quite a few questions in the air from my end.

I have not yet reviewed the results functions in `member-state.ts` I have also not yet reviewed the frontend stuff, but I have to go for now and figured I have given you enough to play with for now 😅

backend/src/services/simulation-service/team-simulator/member-state.ts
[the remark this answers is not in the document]
CarelessParsley: I'd love to know the exact benchmark setup, so I can check if optimizations are speeding them up.

backend/src/services/simulation-service/team-simulator/member-state.ts
CarelessParsley: I actually did think about this optimization and decided not to do it because it was annoying haha

backend/src/services/simulation-service/team-simulator/member-state.ts
CarelessParsley: It's not intentional. Originally I followed the code closely which meant both berry/ing helps, but on Discord people suggested actual number of ings made more sense, which I agreed with, but I didn't update berry lol.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -63,8 +66,30 @@ export class MemberState {
...
+  private totalIngredientHelps = emptyIngredientInventoryFloat();
   private voidHelps = 0;
 
+  // We want to produce histograms showing the distribution of how many helps we
+  // got over the Monte Carlo simulations.  The way that MemberState is
+  // interacted with over the simulations is that we repeatedly run the sim
+  // with the *same* MemberState.  So we cannot have a single counter for
+  // number of helps.  Instead, we need a per-day counter, which we then shove
+  // into an array so we know exactly how much we produced each day.
+  private berryProductionPerDay: number[] = [];
```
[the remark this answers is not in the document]
CarelessParsley: I did this, and now that the arrays are preallocated, I can eliminate currentDayXProduction entirely, as I can just write into the preallocated arrays.

CarelessParsley:
Haven't finished fixing everything yet, resolved comments have changes pushed

backend/src/services/simulation-service/team-simulator/member-state.ts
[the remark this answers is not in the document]
SleepAPI: There's no specific benchmark setup. You can just checkout the develop branch, (rebuild common/backend/frontend), run a couple of team simulations and note the request times of the API responses. Then switch back to your branch and compare.

backend/src/services/simulation-service/team-simulator/member-state.ts
```diff
@@ -63,8 +66,30 @@ export class MemberState {
...
+  // Each element is a Float32Array representing all ingredients for that day
+  private ingredientProductionPerDay: Float32Array[] = [];
```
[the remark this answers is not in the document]
CarelessParsley: From Discord: some ingredient counts will continue to be float because we prefer not to truly simulate ingredient magnet procs, pulling averages for those is better.

CarelessParsley:
So, status report: there's a bunch of nits that still need fixing, but the spill calculation is now finally right, so what's left are the nits + getting the performance back on par.

backend/src/services/simulation-service/team-simulator/member-state/member-state-utils.ts
```diff
@@ -1,4 +1,4 @@
-export function calculateSkillProcDistribution(data: number[]): Record<number, number> {
+export function calculateDistribution(data: number[] | Int16Array | Float32Array): Record<number, number> {
```
SleepAPI, line 1: Found a cool type intersection to make this more generic:
```typescript
type NumericCollection = ArrayLike<number> & Iterable<number>;
...

export function calculateDistribution(data: NumericCollection) ...
```
That allows the function to take in any type of number array.

CarelessParsley:
I did one small optimization which seemed to make a 2% improvement. After it, on my Macbook Pro M4, here are the baseline timings:

```
[LOG] 2025-02-22T03:40:37.037Z: Entered /calculator/team
POST /api/calculator/team 200 11970 - 269.081 ms
[LOG] 2025-02-22T03:40:40.906Z: Entered /calculator/team
POST /api/calculator/team 200 11977 - 243.948 ms
[LOG] 2025-02-22T03:40:41.795Z: Entered /calculator/team
POST /api/calculator/team 200 11995 - 249.444 ms
[LOG] 2025-02-22T03:40:43.957Z: Entered /calculator/team
POST /api/calculator/team 200 11973 - 246.593 ms
[LOG] 2025-02-22T03:40:50.133Z: Entered /calculator/team
POST /api/calculator/team 200 12000 - 271.402 ms
[LOG] 2025-02-22T03:40:52.153Z: Entered /calculator/team
POST /api/calculator/team 200 11993 - 230.779 ms
[LOG] 2025-02-22T03:40:53.054Z: Entered /calculator/team
POST /api/calculator/team 200 12009 - 240.206 ms
[LOG] 2025-02-22T03:40:53.838Z: Entered /calculator/team
POST /api/calculator/team 200 11988 - 255.396 ms
[LOG] 2025-02-22T03:40:55.321Z: Entered /calculator/team
POST /api/calculator/team 200 11977 - 283.978 ms
[LOG] 2025-02-22T03:41:06.382Z: Entered /calculator/team
POST /api/calculator/team 200 11993 - 257.284 ms
```

and timings after my change:

```
[LOG] 2025-02-22T03:39:51.087Z: Entered /calculator/team
POST /api/calculator/team 200 18836 - 279.892 ms
[LOG] 2025-02-22T03:39:53.401Z: Entered /calculator/team
POST /api/calculator/team 200 18826 - 280.910 ms
[LOG] 2025-02-22T03:39:54.469Z: Entered /calculator/team
POST /api/calculator/team 200 18813 - 277.341 ms
[LOG] 2025-02-22T03:39:55.372Z: Entered /calculator/team
POST /api/calculator/team 200 18732 - 280.677 ms
[LOG] 2025-02-22T03:40:58.360Z: Entered /calculator/team
POST /api/calculator/team 200 18895 - 284.208 ms
[LOG] 2025-02-22T03:40:59.132Z: Entered /calculator/team
POST /api/calculator/team 200 18766 - 246.703 ms
[LOG] 2025-02-22T03:40:59.874Z: Entered /calculator/team
POST /api/calculator/team 200 18792 - 244.364 ms
[LOG] 2025-02-22T03:41:00.712Z: Entered /calculator/team
POST /api/calculator/team 200 18839 - 282.582 ms
[LOG] 2025-02-22T03:41:01.499Z: Entered /calculator/team
POST /api/calculator/team 200 18840 - 292.655 ms
[LOG] 2025-02-22T03:41:03.970Z: Entered /calculator/team
POST /api/calculator/team 200 18717 - 263.730 ms
```

So it's not really clear to me that the new logic is that much slower with the optimization.

According to profiling (this is with [a27589e](https://github.com/nerolis-lab/nerolis-lab/pull/548/commits/a27589e305ead2c647874bcd72fa0ecbe759dd8f)):

image

There's probably a few more 1% opportunities here and there, although if you can confirm the code is not /actually/ slower it will be much better to leave the optimizations to a separate PR.

Here's the profile if you want to look: [Timeline%20Recording%201 (1).json](https://github.com/user-attachments/files/18919713/Timeline.20Recording.201.1.json)

backend/src/services/simulation-service/team-simulator/member-state.ts
CarelessParsley: I implemented it. The resulting code is cleaner, so I will keep it. But it does not really help perf lol

backend/src/services/simulation-service/team-simulator/member-state.ts
CarelessParsley: This is dead now that we're precomputing roll thresholds

backend/src/routes/solve-router/solve-router.ts
```diff
@@ -35,12 +35,12 @@ class SolveRouterImpl {
...
-          const body = req.body;
-          const pretty = req.query.pretty;
+          // const name = req.params.name;
+          // const body = req.body;
+          // const pretty = req.query.pretty;
 
-          const data = await solvePool.exec('solveIngredient', [name, body, pretty]);
-          res.json(data);
+          // const data = await solvePool.exec('solveIngredient', [name, body, pretty]);
+          // res.json(data);
```
CarelessParsley, line 43: oopsies

backend/src/utils/random-utils/pre-generated-random.ts
```diff
@@ -0,0 +1,144 @@
+/**
+ * Copyright 2024 Sleep API Authors
```
SleepAPI, line 2: 2025 Neroli's Lab Authors

backend/src/utils/random-utils/pre-generated-random.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import seedrandom from 'seedrandom';
+
+// Fixed seed for consistent random number generation
```
SleepAPI, line 19: We can probably remove a lot of comments in this file. This is giving "// cat" vibes

image

backend/src/utils/random-utils/pre-generated-random.ts
```diff
@@ -0,0 +1,144 @@
...
+const FIXED_SEED = 'seed';
+
+// Maximum value for uint8 (256 values: 0-255)
+const UINT8_MAX = 255;
+
+/**
+ * Global store for pre-generated random numbers
+ */
+export class RandomNumberStore {
```
SleepAPI, line 28: I think it would be quite a lot easier if this exposed a more dedicated singleton instead. For an example see `backend/src/utils/time-utils/time-utils.ts`

We create a class with constructor and any methods we would like to expose, in this case it would have `index` as a private variable and have methods such as `getRandomNumber` and `getRawUint` be public.

Then we export an insstance of that class and in the simulator instead of passing an instance object/interface all the way we would just import the singleton and do `RandomNumberStore.getRandomNumber()`.

What do you think?

SleepAPI: Also that would then mean `instance` and `getInstance` would be removed for example

backend/src/utils/random-utils/pre-generated-random.ts
SleepAPI, line 1: Missing a `pre-generated-random.test.ts` for this

backend/src/services/simulation-service/team-simulator/team-simulator.ts
```diff
@@ -49,8 +52,17 @@ export class TeamSimulator {
...
   private energyDegradeCounter = -1; // -1 so it takes 3 iterations and first degrade is after 10 minutes, then 10 minutes between each
 
-  constructor(params: { settings: TeamSettingsExt; members: TeamMemberExt[]; cookingState?: CookingState }) {
-    const { settings, members, cookingState } = params;
+  constructor(params: {
+    settings: TeamSettingsExt;
+    members: TeamMemberExt[];
+    cookingState?: CookingState;
+    iterations: number;
+    rng?: EnhancedRandom;
```
SleepAPI, line 60: I think both this input param and the line on 65 is unnecessary if we go with dedicated singleton. That way we just decide the seed on startup and access the functionality directly through the singleton import.

backend/src/services/simulation-service/team-simulator/team-simulator.ts
```diff
@@ -253,8 +272,9 @@ export class TeamSimulator {
     const lowestEnergyMembers = sortedMembers.filter((mem) => mem.energy === lowestEnergy);
     const allMembers = this.memberStates;
 
-    const targetGroup = RandomUtils.roll(chanceTargetLowest) ? lowestEnergyMembers : allMembers;
-    return [RandomUtils.randomElement(targetGroup)].filter((member): member is MemberState => member !== undefined);
+    const targetGroup = this.rng() < chanceTargetLowest ? lowestEnergyMembers : allMembers;
+    const randomIndex = Math.floor(this.rng() * targetGroup.length);
```
SleepAPI, line 276: I would probably make a helper method for this on the RandomNumberStore to make this more readable

backend/src/utils/random-utils/pre-generated-random.ts
SleepAPI, line 1: I would probably name the file and class similar

SleepAPI (changes requested):

SleepAPI:
Another note; you'll need to bump version.ts again since it was bumped in develop

backend/src/services/simulation-service/team-simulator/team-simulator.ts
CarelessParsley: My main hesitation with a global singleton EnhancedRandom is when we do testing, we will have to remember to reset it before every test, or we can end up with some nondeterministic tests if you run them in the wrong order. It seemed more explicit and easier to control to just pass them around. I also am not sure how it interacts with multithreaded workers (does each worker get its own copy of all heap objects? I'm fuzzy how this works.)

backend/src/utils/random-utils/pre-generated-random.ts
CarelessParsley: I'm happy to do the initialization for whatever singletons we have at module load. We need to figure out if we want the index to be a global though (see other thread).

backend/src/services/simulation-service/team-simulator/team-simulator.ts
SleepAPI: Worker threads can share memory, but by default don't. So if you create a new worker then as far as I know it should create a new singleton with the same fixed seed as other workers. So I think we should be good here?

As far as testing goes we could add a reset method to the class that sets index to 0. Then we could create a setup file, maybe under bun folder and probably rename bun to vitest since the tests don't use bun anymore nayway. This setup file would include a beforeEach call to the reset function. Then in our vitest.config.ts we add that setup file to `setupFiles`. That will automatically reset index to 0 for all tests, before each test is run

SleepAPI (approved):
Massive change, amazing work! 🎉

backend/src/services/simulation-service/team-simulator/cooking-state/cooking-state.ts
```diff
@@ -56,11 +58,12 @@ export class CookingState {
   private currentSaladStockpile: IngredientIndexToFloatAmount;
   private currentDessertStockpile: IngredientIndexToFloatAmount;
 
-  constructor(settings: TeamSettingsExt, userRecipes: UserRecipes) {
+  constructor(settings: TeamSettingsExt, userRecipes: UserRecipes, rng: PreGeneratedRandom | null = null) {
```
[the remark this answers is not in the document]
CarelessParsley: Simple enough to fix all the call sites.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -63,7 +69,19 @@ export class MemberState {
   private helpsSinceLastCook = 0;
   private totalAverageHelps = 0;
```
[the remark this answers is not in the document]
CarelessParsley: let's see if I can get Cursor to update all the mocks lol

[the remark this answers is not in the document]
CarelessParsley: This is at least a little nontrivial, because it is used for both averageHelps and totalHelps (for simple) returns. So we'd have to kill a few more API calls. Did you expect this?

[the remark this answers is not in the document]
CarelessParsley: It seems to be used in places like backend/src/services/solve/utils/solve-utils.ts . Do you propose that the averageHelps quantity instead be replaced with the totalHelps quantity?

PR #554, opened 2025-03-01
test: adding sneaky snacking functionality to common/backend INCOMPLETE

jlau3:


backend/src/controllers/calculator/production.controller.ts
```diff
@@ -122,6 +122,7 @@ export default class ProductionController extends Controller {
             camp
           }),
           nature: getNature(member.nature),
+          isSneakySnacking: member.isSneakySnacking,
```
SleepAPI, line 125: I don't think this is available in `PokemonInstanceIdentity`? 

But also if you're updating the backend API to expect sneaky snacking the frontend will have to send it too (in `team-service.ts`

backend/src/database/migration/migrations/008_migration.js
```diff
@@ -0,0 +1,15 @@
+const Tables = Object.freeze({
+  TeamMember: 'team_member'
+});
+
+export async function up(knex) {
+  await knex.schema.alterTable(Tables.TeamMember, (table) => {
+    table.string('is_sneaky_snacking', 1024).notNullable().defaultTo('false');
```
SleepAPI, line 7: Let's do table.boolean, notNullable, defaultTo false. Perhaps we also just call it `sneaky_snacking`?

backend/src/routes/team-router/team-router.ts
```diff
@@ -81,7 +81,7 @@ class TeamRouterImpl {
       validateAuthHeader,
       async (
         req: Request<
-          { teamIndex: string; memberIndex: string },
+          { teamIndex: string; memberIndex: string; isSneakySnacking: string },
```
SleepAPI, line 84: So these are the query params, currently the upsert team member is called like:
`api/team/0/member/0` if updating the first member in first team. We probably don't want to add the sneaky snacking as query on top of that.

Rather we should put it in the request, which would be `UpsertTeamMemberRequest` 

Maybe something like:
```typescript
export type UpsertTeamMemberRequest = PokemonInstanceWithMeta & { sneakySnacking: boolean }
```

backend/src/database/dao/team/team-member-dao.ts
```diff
@@ -7,6 +7,7 @@ const DBTeamMemberSchema = Type.Composite([
   Type.Object({
     fk_team_id: Type.Number(),
     fk_pokemon_id: Type.Number(),
+    is_sneaky_snacking: Type.Boolean(),
```
SleepAPI, line 10: If we rename we should do it here too

backend/src/services/api-service/team/team-service.ts
```diff
@@ -94,12 +95,18 @@ export async function upsertTeamMember(params: {
...
   const updatedMemberMeta = await TeamMemberDAO.upsert({
-    updated: { fk_pokemon_id: upsertedMember.id, fk_team_id: updatedTeam.id, member_index: memberIndex },
-    filter: { fk_team_id: updatedTeam.id, member_index: memberIndex }
+    updated: {
+      fk_pokemon_id: upsertedMember.id,
+      fk_team_id: updatedTeam.id,
+      is_sneaky_snacking: isSneakySnacking,
+      member_index: memberIndex
+    },
+    filter: { fk_team_id: updatedTeam.id, is_sneaky_snacking: isSneakySnacking, member_index: memberIndex }
```
SleepAPI, line 104: We shouldn't include isSneakySnacking in the filter because that means when we try to set a value that was false to true we will only search for entries in the database that are already true (which will never exist).

backend/src/database/migration/migrations/008_migration.js
jlau3: just a habit I picked up for boolean values. I saw you folks already had sneakySnack here: https://github.com/nerolis-lab/nerolis-lab/blob/develop/common/src/api/team/calculate/team-calculate.ts#L46
so thought I should differentiate

backend/src/controllers/calculator/production.controller.ts
jlau3: oops that's a mistake, forgot to push that file

backend/src/database/migration/migrations/008_migration.js
SleepAPI: That's an array of berries that were actually sneaky snacked. We should be able to tell fro context (berry array vs boolean) what it is imo

backend/src/routes/team-router/team-router.ts
```diff
@@ -101,6 +103,7 @@ class TeamRouterImpl {
           const updatedMember = await controller.upsertMember({
             teamIndex: +teamIndex,
             memberIndex: +memberIndex,
+            isSneakySnacking: queryAsBoolean(isSneakySnacking),
             request: req.body,
```
SleepAPI, line 107: If we're passing the entire body anyway I don't think we have to specifically break out sneaky snacking here, that essentially just means we're passing it twice

SleepAPI (changes requested):
Since we're changing the API it's probably safest to also update the version in `version.ts` which invalidates the frontend cache

SleepAPI:
Migration 008 was just added in develop, so I recommend renaming your migration to 009 before rebase to avoid conflicts

PR #579, opened 2025-03-27
feat: implement sneaky snacking

jlau3:


SleepAPI:
The commits seem to have gotten weird with this one. There's 3 duplicates and 3 merge commits. We have a LINEAR_HISTORY.md now if that helps

backend/src/database/migration/migrations/010_migration.js
```diff
@@ -0,0 +1,15 @@
+const Tables = Object.freeze({
+  TeamMember: 'team_member'
+});
+
+export async function up(knex) {
+  await knex.schema.alterTable(Tables.TeamMember, (table) => {
+    table.boolean('sneaky_snacking', 1024).notNullable().defaultTo('false');
```
SleepAPI, line 7: No need to specify length of this to 1024, booleans in mysql are just 1 or 0. You can remove the length param

backend/src/services/api-service/team/team-service.ts
```diff
@@ -94,8 +94,13 @@ export async function upsertTeamMember(params: {
...
   const updatedMemberMeta = await TeamMemberDAO.upsert({
-    updated: { fk_pokemon_id: upsertedMember.id, fk_team_id: updatedTeam.id, member_index: memberIndex },
-    filter: { fk_team_id: updatedTeam.id, member_index: memberIndex }
+    updated: {
+      fk_pokemon_id: upsertedMember.id,
+      fk_team_id: updatedTeam.id,
+      sneaky_snacking: request.sneakySnacking,
+      member_index: memberIndex
+    },
+    filter: { fk_team_id: updatedTeam.id, sneaky_snacking: request.sneakySnacking, member_index: memberIndex }
```
SleepAPI, line 103: The update looks good, but let's remove sneaky snacking from the filter. The filter is used to find records in the database to update. If you search for records matching your incoming request.sneakySnacking you'll never find a result is you're trying to update sneaky snacking from false to true (since it would search for records already being true for this member index)

SleepAPI (changes requested):
~~I haven't gone through it all yet, but I'll continue on desktop later. This is just some initial findings. Excited to get sneaky snacking in, it's a highly requested feature~~

EDIT: review completed

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -177,8 +179,15 @@ export class MemberState {
 
     this.nextHelp = this.fullDayDuration; // set to 1440, first start of day subtracts 1440
 
-    this.skillPercentage = TeamSimulatorUtils.calculateSkillPercentage(member);
-    this.ingredientPercentage = TeamSimulatorUtils.calculateIngredientPercentage(member);
+    this.sneakySnacking = member.settings.sneakySnacking;
+
+    if (this.sneakySnacking === true) {
```
SleepAPI, line 184: Probably a little verbose, also since sneakySnacking is already a boolean we can check it directly:
```typescript
this.skillPercentage = this.sneakySnacking ? 0 : TeamSimulatorUtils.calculateSkillPercentage(member);
this.ingredientPercentage = this.sneakySnacking ? 0 : TeamSimulatorUtils.calculateIngredientPercentage(member);
```

backend/src/services/simulation-service/team-simulator/team-simulator.test.ts
SleepAPI, line 1: I would love to see a test added here with maybe like 10 iterations and then assert that 0 skill procs have happened and no ingredients have been gathered. That should verify edge cases like pity procs as well. Could verify that the amount of help x berries per drop matches, since every help should be berries

backend/src/services/simulation-service/team-simulator/skill-state/skill-state.test.ts
```diff
@@ -147,4 +148,21 @@ describe('SkillState', () => {
...
   });
+
+  it('should skip all skill activation when sneakySnacking is true', () => {
+    mockMemberState.sneakySnacking = true;
+
+    const activations: TeamSkillActivation[] = skillState.attemptSkill();
+    expect(activations).toEqual([]);
+  });
+
+  it('should allow pity proc activation when sneakySnacking is false', () => {
```
SleepAPI, line 159: I think this test already exists?

backend/src/services/simulation-service/team-simulator/skill-state/skill-state.ts
```diff
@@ -87,6 +87,11 @@ export class SkillState {
   // // TODO: apparently returning early here makes the team sim insanely fast, so skill handling is slower than expected
   public attemptSkill(): TeamSkillActivation[] {
     const activations: TeamSkillActivation[] = [];
+
+    if (this.memberState.sneakySnacking === true) {
+      return activations;
```
SleepAPI, line 92: Could probably move this above the activations array even and we can just do 
```typescript
if(this.memberState.sneakySnacking
```

backend/src/services/simulation-service/team-simulator/team-simulator.ts
```diff
@@ -42,6 +42,7 @@ export class TeamSimulator {
   private memberStates: MemberState[] = [];
   private memberStatesWithoutFillers: MemberState[] = [];
   private cookingState?: CookingState = undefined;
+  private sneakySnacking = false;
```
SleepAPI, line 45: I don't think this is needed, and also I don't think we have to pass `sneakySnacking` to MemberState since we're already passing settings

common/src/domain/optimal/optimal.ts
```diff
@@ -20,11 +20,12 @@ export interface Optimal {
   nature: Nature;
   carrySize: number;
   skillLevel: number;
+  sneakySnacking: boolean;
   ribbon: number;
 }
 
 class OptimalImpl {
-  public berry(pokemon: Pokemon, ribbon?: number): Optimal {
+  public berry(pokemon: Pokemon, sneakySnacking?: boolean, ribbon?: number): Optimal {
```
SleepAPI, line 28: These are starting to become a little annoying now, how about we params wrap the parameters?
```typescript
public berry(params: {pokemon: Pokemon, sneakySnacking?: boolean, ribbon?: number})
```
That way we can do
```typescript
Optimal.berry({pokemon: SNEASEL})
```
instead of
```typescript
// idk what false and 3 refer to looking at this
Optimal.berry(pokemon.SNEASEL, false, 3)
```

SleepAPI: this goes for the other methods here too like skill and ingredient

frontend/src/services/team/team-service.ts
```diff
@@ -189,17 +189,17 @@ class TeamServiceImpl {
...
     const berrySetup: PokemonInstanceIdentity = PokemonInstanceUtils.toPokemonInstanceIdentity({
       ...currentMember,
-      ...Optimal.berry(currentMember.pokemon, currentMember.ribbon),
+      ...Optimal.berry(currentMember.pokemon, currentMember.sneakySnacking, currentMember.ribbon),
       externalId: uuid.v4()
     })
     const ingredientSetup: PokemonInstanceIdentity = PokemonInstanceUtils.toPokemonInstanceIdentity({
       ...currentMember,
-      ...Optimal.ingredient(currentMember.pokemon, currentMember.ribbon),
+      ...Optimal.ingredient(currentMember.pokemon, currentMember.sneakySnacking, currentMember.ribbon),
```
SleepAPI, line 197: For ingredient and skill I don't think we should copy sneakySnacking here, we can copy it for berry optimal setup though. But it's more clear to the user I think to show 0 for skill and ingredient than 100 score.

package.json
```diff
@@ -34,5 +34,8 @@
     "semantic-release": "^24.2.0",
     "typescript": "^5.7.2",
     "typescript-eslint": "^8.17.0"
+  },
+  "dependencies": {
```
SleepAPI, line 38: I think this was a mistake somehow, this is only used in backend and it's already part of backend package.json

common/src/domain/instance/pokemon-instance.ts
```diff
@@ -29,10 +29,11 @@ export interface PokemonInstanceMeta {
   gender: PokemonGender;
   name: string;
 }
-export type PokemonInstanceWithMeta = PokemonInstance & PokemonInstanceMeta;
+export type PokemonInstanceWithMeta = PokemonInstance & PokemonInstanceMeta & { sneakySnacking: boolean };
```
SleepAPI, line 32: Hmmm thinking about this more, I'm not sure this works well. This is essentially tied to a PokemonInstance which gets inserted into the database pokemon table. But the same PokemonInstance can be added into multiple teams or even multiple times in the same team. The sneakySnacking toggle is very much team member specific.

In the frontend in the `TeamInstance` interface we have a members array like:
```typescript
  members: (string | undefined)[]
  memberIvs: Record<string, PerformanceDetails | undefined>
```
The string here refer to external ids which are used to look up the Pokémon, so members array could contain the same externalId 5 times technically. I think it should be changed to a different type than string now. It should probably be a new member interface, maybe call it TeamMemberIdentity or something more descriptive, that is pokemonExternalId: string and sneakySnacking: boolean.

backend/src/services/user-service/user-pokemon-service/user-pokemon.ts
```diff
@@ -20,6 +20,7 @@ export async function getSavedPokemon(user: DBUser): Promise<PokemonInstanceWith
     skillLevel: pkmn.skill_level,
     nature: pkmn.nature,
     subskills: PokemonDAO.filterFilledSubskills(pkmn),
+    sneakySnacking: false,
```
SleepAPI, line 23: Yeah this is a side effect of including sneakySnacking on PokemonInstanceWithMeta. It doesn't make sense here, sneakySnacking is unrelated to a saved Pokémon. It should probably not be part of PokemonInstanceWithMeta

SleepAPI (changes requested):

backend/src/services/simulation-service/team-simulator/team-simulator.ts
jlau3: sneakySnacking isn't a part of settings, but I can include it in there, and reference that instead

jlau3: Actually giving this another thought, do we want to include it in settings? TeamSettingsExt is for the team at large while we wanna be able to sneaky snack individually

SleepAPI: You're right, it should be part of the member we're passing to MemberState. member contains settings which is TeamMemberSettingsExt related to that specific member, which should contain sneakySnacking data.

PR #583, opened 2025-04-03
fix: replace cached sleep score with store value

SleepAPI:


frontend/src/components/calculator/team-settings/team-settings.vue
```diff
@@ -30,12 +30,16 @@
       >
         <v-avatar size="48">
           <v-progress-circular
-            :model-value="sleepScore({ bedtime, wakeup })"
+            :model-value="
+              sleepScore({ bedtime: teamStore.getCurrentTeam.bedtime, wakeup: teamStore.getCurrentTeam.wakeup })
```
tindomerel8, line 34: If you want all this to look tidy, you could make bedtime and wakeup calculated fields that just return teamStore.getCurrentTeam.bedtime, then if they ever need to change, you don't need to change them in 6+ places. But nbd

tindomerel8 (approved):

frontend/src/components/calculator/team-settings/team-settings.vue
SleepAPI: I think you're right, I considered this, but was cautious with mutability. `model-value` is used both as the shown number (read) and as the value to update (write). Computed properties are read-only I believe, so I wasn't sure if that'd work.

I'll give it a try

SleepAPI: Actually I dont think thats true, model-value is read only I think, v-model is both.
I might just have to be careful with the update functionality further down

tindomerel8: Oh, yeah, you'd need to make a separate setter for it in that case. May still come out cleaner that way, but not quite as seamless as I first thought when suggesting it.

SleepAPI: Solved it, made all read values use a computed property, kept the write as is

frontend/src/components/calculator/team-settings/team-settings.vue
```diff
@@ -125,11 +125,13 @@
...
           <v-col class="flex-center">
             <v-btn color="primary" aria-label="bedtime button" @click="isBedtimeOpen = true"
-              >Bedtime: {{ bedtime }}</v-btn
+              >Bedtime: {{ teamStore.getCurrentTeam.bedtime }}</v-btn
             >
           </v-col>
           <v-col class="flex-center">
-            <v-btn color="primary" aria-label="wakeup button" @click="isWakeupOpen = true">Wakeup: {{ wakeup }}</v-btn>
+            <v-btn color="primary" aria-label="wakeup button" @click="isWakeupOpen = true"
+              >Wakeup: {{ teamStore.getCurrentTeam.wakeup }}</v-btn
```
tindomerel8, line 133: These two could use the computed value, right?

SleepAPI: Huh Idk how I missed this, good catch, fixed and updated

tindomerel8 (approved):

PR #588, opened 2025-04-18
Update 2.7.0

tindomerel8:


common/src/domain/mainskill/mainskills/dream-shard-magnet-s.ts
SleepAPI, line 1: I get this change. I think we need to go away from max skill level. We use it for optimal mons in some places

SleepAPI (approved):

SleepAPI (approved):

PR #590, opened 2025-04-24
fix(auth)!: patreon supporter status links correctly

SleepAPI:
This commit rewrites the entire oauth flow, including the Google implementation.

This adds Discord and Patreon oauth provider alternatives.

This adds linking of additional providers.

This adds tokens of appreciation for site supporters.

BREAKING CHANGE: Google client needs a redirect uri now and the API has been updated with new required headers.

### DIFF
--- a/frontend/src/services/login/patreon-service.ts
+++ b/frontend/src/services/login/patreon-service.ts
@@ -3,7 +3,8 @@ import type { RouteLocationNormalizedLoaded } from 'vue-router'
 
 export const PATREON_REDIRECT_URI = `${window.location.origin}/patreon`
 
-export function getPatreonAuthCode(currentRoute: RouteLocationNormalizedLoaded, scope = 'identity[email]') {
+export function getPatreonAuthCode(currentRoute: RouteLocationNormalizedLoaded) {
+  const scope = 'identity identity[email]'
   const params = new URLSearchParams({
     client_id: import.meta.env.VITE_PATREON_CLIENT_ID,
     redirect_uri: PATREON_REDIRECT_URI,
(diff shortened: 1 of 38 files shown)


frontend/src/services/login/patreon-service.ts
```diff
@@ -3,7 +3,7 @@ import type { RouteLocationNormalizedLoaded } from 'vue-router'
 
 export const PATREON_REDIRECT_URI = `${window.location.origin}/patreon`
 
-export function getPatreonAuthCode(currentRoute: RouteLocationNormalizedLoaded, scope = 'identity[email]') {
+export function getPatreonAuthCode(currentRoute: RouteLocationNormalizedLoaded, scope = 'identity identity[email]') {
```
tindomerel8, line 6: Scope here seems like it should be a constant. I don't see anywhere where we have the need to call this method with alternative scopes.

frontend/src/services/utils/time-utils.ts
```diff
@@ -44,6 +44,10 @@ class TimeUtilsImpl {
     return `${hours !== '00' ? `${hours}h ` : ''}${minutes}m ${seconds}s`
   }
 
+  public extractDate(isoString: string): string {
+    return isoString.split('T')[0]
```
tindomerel8, line 48: If we wanted this to work for the user's region, we could use this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

For example: 
(date object).toLocaleDateString(undefined, { dateStyle: "long" }) could produce the following outputs depending on the user's browser settings:
"19. Dezember 2012" (de-DE)
 "December 19, 2012" (en-US)

tindomerel8 (dismissed):

frontend/src/services/utils/time-utils.ts
SleepAPI: Fixing this!

tindomerel8 (approved):

PR #642, opened 2025-06-03
Sneaky Snacking

3geek14:
Add a `sneakySnacking` boolean to `TeamMemberSettings`. In `member-state.ts`, if the member is sneaky snacking, don't roll ingredients for normal helps, and never activate skills.

Ran the benchmarking tests for the team simulator. The results are consistent, and the runs took about the same amount of time.

### DIFF
--- a/backend/src/controllers/calculator/production.controller.ts
+++ b/backend/src/controllers/calculator/production.controller.ts
@@ -213,7 +213,8 @@ export default class ProductionController {
           nature: getNature(member.nature),
           skillLevel: Math.min(member.skillLevel, pokemon.skill.maxLevel),
           subskills,
-          externalId: member.externalId
+          externalId: member.externalId,
+          sneakySnacking: member.sneakySnacking
         }
       });
     }
--- a/backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
+++ b/backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
@@ -316,6 +316,10 @@ export class MemberState {
     return this.member.settings.externalId;
   }
 
+  get isSneakySnacking() {
+    return this.member.settings.sneakySnacking;
+  }
+
   public wakeUp() {
     const nrOfErb = TeamSimulatorUtils.countMembersWithSubskill(this.team, subskill.ENERGY_RECOVERY_BONUS.name);
     const sleepInfo: SleepInfo = {
@@ -512,6 +516,11 @@ export class MemberState {
       return [];
     }
 
+    if (this.isSneakySnacking) {
+      this.attemptSneakySnackingHelp('day');
+      return [];
+    }
+
     this.addBerriesAndIngredientsForHelp('day');
     return this.skillState.attemptSkill();
   }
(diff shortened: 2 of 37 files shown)


backend/src/controllers/calculator/production.controller.ts
```diff
@@ -209,7 +209,8 @@ export default class ProductionController {
           nature: getNature(member.nature),
           skillLevel: member.skillLevel,
           subskills,
-          externalId: member.externalId
+          externalId: member.externalId,
+          sneakySnacking: false // default team members to not sneaky snacking
```
SleepAPI, line 213: Do we want to prepare this for the frontend part? I'm thinking after this change the backend should be done and we would only need to add the frontend UI / send correct value. 

So instead of defaulting sneakySnacking here I'm thinking perhaps we update the API to take in sneakySnacking for member settings and then for now the frontend can send default false.

Thoughts?

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -759,3 +806,5 @@ export class MemberState {
     return this.frequency0;
   }
 }
+
+export type HelpPeriod = 'day' | 'night';
```
SleepAPI, line 810: Minor and feel free to ignore, but I tend to declare types at the top of a file

webbywisp (dismissed):

backend/src/controllers/calculator/production.controller.ts
3geek14: Yeah, I think it makes sense for `parseTeamMembers` to accept sneaky snacking as an input. That means that `PokemonInstanceBase` needs to track sneaky snacking as well, or else I need to add another parameter. I'm not sure quite how deeply we want to store this: I think the Pokemon Box shouldn't track whether a mon is sneaky snacking, because the same mon on different teams might sneaky snack and collect regularly. But saved teams should remember whether a mon is sneaky snacking, so that things don't change between devices. I'm not sure what distinction we have between Pokemon and Team Members in the database.

* I think `getSavedPokemon(user: DBUser)` can default to not sneaky snacking, because I think this is getting a list of mons in the box.
* `findTeamsWithMembers` seems like it should know about sneaky snacking, so the `member: DBPokemon` it gets from `PokemonDAO.get({ id: memberData.fk_pokemon_id })` has to know about sneaky snacking. But I think that's the same as the type we use for mons in the box, so I'm not confident it's safe for me to add it. So I'm going to default to `false` here, and wait to hear back.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
3geek14: Done. And noted for my refactor PR.

backend/src/services/user-service/user-pokemon-service/user-pokemon.ts
```diff
@@ -20,7 +20,8 @@ export async function getSavedPokemon(user: DBUser): Promise<PokemonInstanceWith
     skillLevel: pkmn.skill_level,
     nature: pkmn.nature,
     subskills: PokemonDAO.filterFilledSubskills(pkmn),
-    ingredients: PokemonDAO.filterChosenIngredientList(pkmn)
+    ingredients: PokemonDAO.filterChosenIngredientList(pkmn),
+    sneakySnacking: false // default team members to not sneaky snacking
```
SleepAPI, line 24: These aren't necessarily team members. This returns all the pokemon that a user has saved to pokebox.

webbywisp (changes requested):

backend/src/services/user-service/user-pokemon-service/user-pokemon.ts
3geek14: So defaulting to false makes sense for box mons?

webbywisp (dismissed):

backend/src/database/dao/team/team-dao.ts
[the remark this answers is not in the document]
3geek14: Looks like the file probably shouldn't be here. Good catch!

tindomerel8 (approved):

PR #649, opened 2025-06-07
feat(iv graph) Use mon's skill level to compare

3geek14:
Rather than using max skill levels for optimal mons, use the same skill level as the user entered.

Sleep API will still use the max skill level, because that's also used for the tier lists, and I think it makes sense for tier lists to use fully-invested mons.


3geek14:
Blastoise with level 3 main skill gets 100% contribution now.

image

SleepAPI (approved):
Awesome!

PR #669, opened 2025-06-17
feat(ingredient draw): use updated probabilities

3geek14:
New data from RaenonX has more reliable estimates for the probabilities of each outcome for these skills. These data suggest the ingredients might not be weighted equally, which requires a refactor.

### DIFF
--- a/common/src/utils/mainskill-utils/mainskill-utils.ts
+++ b/common/src/utils/mainskill-utils/mainskill-utils.ts
@@ -4,6 +4,10 @@ export function rollToOutput<TSkillOutput extends string | number | symbol>(
 ): TSkillOutput {
   const allOutputs: TSkillOutput[] = Object.keys(probabilities) as TSkillOutput[];
 
+  if (allOutputs.length === 0) {
+    throw new Error('No skill outputs provided');
+  }
+
   let totalProbability = 0;
   for (const output of allOutputs) {
     totalProbability += probabilities[output];
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -119,7 +119,6 @@ export default typescriptEslint.config(
         }
       ],
       '@typescript-eslint/no-unused-vars': [
-        // this seems to apply to backend but not to common???
         'error',
         {
           argsIgnorePattern: '^_',
--- a/backend/src/services/calculator/skill/activation/__snapshots__/skill-activation.test.ts.snap
+++ b/backend/src/services/calculator/skill/activation/__snapshots__/skill-activation.test.ts.snap
@@ -935,14 +935,14 @@ exports[`activateMetronome > shall use the correct metronomeFactor for each skil
     "nrOfHelpsToActivate": 2,
     "skill": IngredientDrawSHyperCutterImpl {
       "HyperCutterProbabilities": {
-        "Corn": 19,
-        "Corn (L)": 4,
-        "Oil": 20,
-        "Oil (L)": 3,
-        "Potato": 22,
-        "Potato (L)": 6,
-        "Tomato": 22,
-        "Tomato (L)": 4,
+        "Corn": 20.9,
+        "Corn (L)": 4.1,
+        "Oil": 20.9,
+        "Oil (L)": 4.1,
+        "Potato": 20.9,
+        "Potato (L)": 4.1,
+        "Tomato": 20.9,
+        "Tomato (L)": 4.1,
       },
       "HyperCutterResultConfig": {
         "Corn": {
@@ -1109,12 +1109,12 @@ exports[`activateMetronome > shall use the correct metronomeFactor for each skil
         5843,
       ],
       "SuperLuckProbabilities": {
-        "Coffee": 22,
-        "Dream Shards (L)": 3,
-        "Dream Shards (S)": 13,
-        "Mushroom": 20,
-        "Sausage": 22,
-        "Soybeans": 20,
+        "Coffee": 21.3,
+        "Dream Shards (L)": 2.7,
+        "Dream Shards (S)": 12.1,
+        "Mushroom": 21.3,
+        "Sausage": 21.3,
+        "Soybeans": 21.3,
       },
       "SuperLuckResultConfig": {
         "Coffee": {
--- a/common/src/types/mainskill/mainskills/ingredient-draw-s-hyper-cutter.test.ts
+++ b/common/src/types/mainskill/mainskills/ingredient-draw-s-hyper-cutter.test.ts
@@ -48,6 +48,6 @@ describe('IngredientDrawSHyperCutter', () => {
     for (const key of Object.keys(IngredientDrawSHyperCutter.HyperCutterProbabilities) as HyperCutterOutput[]) {
       sum += IngredientDrawSHyperCutter.HyperCutterProbabilities[key];
     }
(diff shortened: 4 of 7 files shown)


backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/ingredient-draw-s-hyper-cutter-effect.ts
```diff
@@ -15,26 +14,18 @@ export class IngredientDrawSHyperCutterEffect implements SkillEffect {
     const skill = IngredientDrawSHyperCutter;
 
     const roll = skillState.rng();
-    const rolledIngredient: Ingredient = skillState.rng.randomElement(
-      IngredientDrawSHyperCutter.ingredientDrawIngredients
-    );
+    const result = skill.rollToResult(roll);
```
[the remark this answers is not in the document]
SleepAPI: I guess this applies to Super Luck as well

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/ingredient-draw-s-super-luck-effect.ts
```diff
@@ -16,57 +15,40 @@ export class IngredientDrawSSuperLuckEffect implements SkillEffect {
     const nrOfIngredients = skillState.skillAmount(skill.activations.ingredients);
 
     const roll = skillState.rng();
-
+    const result = skill.rollToResult(roll);
```
SleepAPI, line 18: Same applies here regarding unifying the logic in `rollToResult`, then we could probably also skip the unit checks below, since the function would just return both regular and crit amounts.

common/src/domain/mainskill/mainskills/ingredient-draw-s-hyper-cutter.test.ts
SleepAPI, line 1: There are a lot of new functions in the hyper cutter and super luck classes that I think we could be testing

common/src/domain/mainskill/mainskills/ingredient-draw-s-hyper-cutter.ts
```diff
@@ -24,7 +22,87 @@ export const IngredientDrawSHyperCutter = new (class extends ModifiedMainskill {
     }
   };
 
-  get ingredientDrawIngredients(): Ingredient[] {
-    return [SOFT_POTATO, PURE_OIL, SNOOZY_TOMATO, GREENGRASS_CORN];
+  ingredientDrawProbability(result: HyperCutterResult): number {
```
SleepAPI, line 25: This feels like a lot of chained checks. You already created `HyperCutterResult`, could we perhaps create a local config-ish object, like a:
```typescript
const HyperCutterResultConfig: Record<HyperCutterResult, number> = {
  'Potato': 22,
  ...etc
}
```

SleepAPI: It could probably even return an object like:
```typescript
'Potato': { probability: 22, amount: this.activations.ingredients.amount, critAmount: 0 },
... etc,
'Potato (L)': { probability: 22, amount:0, critAmount: this.activations.ingredients.critAmount },
...etc
```
and perhaps that could simplify the `resultToX` functionalities.

common/src/domain/mainskill/mainskills/ingredient-draw-s-hyper-cutter.ts
```diff
@@ -24,7 +22,87 @@ export const IngredientDrawSHyperCutter = new (class extends ModifiedMainskill {
...
+                  ? 4
+                  : result == 'Corn (L)'
+                    ? 4
+                    : 0;
+  }
+
+  rollToResult(roll: number): HyperCutterResult {
+    // roll should be in the interval [0,1).
+
+    let cumulativeProbability = this.ingredientDrawProbability('Potato');
```
SleepAPI, line 48: minor: this definitely works and is readable so I'm fine with this. Having said that, you could do something like:
```typescript
// TODO: I guess order doesn't matter? Perhaps we could just iterate over the Object.keys(config) if we go with the Record config solution
const order: HyperCutterResult[] = [
  'Potato', 'Oil', 'Tomato', 'Corn',
  'Potato (L)', 'Oil (L)', 'Tomato (L)', 'Corn (L)'
];

let cumulativeProbability = 0;
for (const result of order) {
  cumulativeProbability += this.ingredientDrawProbability(result);
  if (roll * 100 < cumulativeProbability) {
    return result;
  }
```

common/src/domain/mainskill/mainskills/ingredient-draw-s-hyper-cutter.ts
3geek14: I'm doing two records, so the probability can be separate. That's letting me more easily extract some the `rollToOutput` function, and also I think it makes the code more readable.

3geek14 (commented):
Thanks for the feedback! I think this looks a lot cleaner now.

common/src/domain/mainskill/mainskills/ingredient-draw-s-hyper-cutter.test.ts
3geek14: There are far fewer functions now, and I think the important one to test was `rollToResult`. That's now a util function, and I've got a handful of tests for it, but I'm happy to add more if you've got ideas. I also think it's good to test that the probabilities add to 100%, to catch a typo I had earlier where the Super Luck probabilities only added to 99. (I'm not sure if that was a typo I made during this refactor or if I'd previously uploaded it, but it seems like an easy typo to make.)

common/src/utils/mainskill-utils/mainskill-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  let cumulativeProbability = 0;
+  for (const output of allOutputs) {
+    cumulativeProbability += probabilities[output];
+    if (roll * totalProbability < cumulativeProbability) {
+      return output;
+    }
+  }
+
+  // In case of rounding issues, return some default value.
```
SleepAPI, line 20: that's assuming allOutputs has length > 0, otherwise this would return undefined. Minor thing though since hopefully we'd never use it with empty probabilities

eslint.config.mjs
```diff
@@ -98,6 +98,14 @@ export default typescriptEslint.config(
           prefer: 'type-imports'
         }
       ],
+      '@typescript-eslint/no-unused-vars': [
+        // this seems to apply to backend but not to common???
+        'error',
+        {
+          argsIgnorePattern: '^_',
```
SleepAPI, line 105: Didn't know about this one, that's very cool!

eslint.config.mjs
```diff
@@ -98,6 +98,14 @@ export default typescriptEslint.config(
           prefer: 'type-imports'
         }
       ],
+      '@typescript-eslint/no-unused-vars': [
+        // this seems to apply to backend but not to common???
```
SleepAPI, line 102: This seems forgotten, there are multiple removals of ignore comments suggesting this should be working in common now

SleepAPI (approved):

SleepAPI (approved):

PR #732, opened 2025-07-14
Base pokemon util function

3geek14:
Add a util function for creating Pokemon. The primary motivation for this is to allow us to specify A/B(/C) ingredients for Pokemon, rather than needing to specify amounts at each level. Darkrai is a special case and doesn't follow the normal rules, so the function should allow putting in the full `IngredientSet[]` versions as well.

Recently, @matt063481 [suggested on Discord](https://discord.com/channels/1300099710996058252/1332444692188762112/1394064394551427293) adding a Pokedex page to the site. If we have this util function, I think we'd be able to more easily set evolution data for each Pokemon. The util function can set new fields `previousForm` and `nextForms` to `undefined` and `[]`, and the existing evolution utilities can mutate the inputs to store names of mons in the evolutionary line.

### DIFF
--- a/common/src/types/pokemon/all-pokemon.ts
+++ b/common/src/types/pokemon/all-pokemon.ts
@@ -1,3 +1,4 @@
+import { basePokemon } from '../../utils';
 import { toSeconds } from '../../utils/time-utils/frequency-utils';
 import { WIKI } from '../berry/berries';
 import { GENDER_UNKNOWN } from '../gender';
@@ -15,7 +16,7 @@ import {
 import { ChargeStrengthMBadDreams } from '../mainskill/mainskills/charge-strength-m-bad-dreams';
 import type { Pokemon } from './pokemon';
 
-export const DARKRAI: Pokemon = {
+export const DARKRAI: Pokemon = basePokemon({
   name: 'DARKRAI',
   displayName: 'Darkrai',
   pokedexNumber: 491,
@@ -61,7 +62,7 @@ export const DARKRAI: Pokemon = {
     { amount: 4, ingredient: ROUSING_COFFEE }
   ],
   skill: ChargeStrengthMBadDreams
-};
+});
 
 export const OPTIMAL_ALL_SPECIALISTS: Pokemon[] = [DARKRAI];
 export const INFERIOR_ALL_SPECIALISTS: Pokemon[] = [];
--- a/common/src/types/pokemon/all-pokemon.ts
+++ b/common/src/types/pokemon/all-pokemon.ts
@@ -17,7 +17,6 @@ import { ChargeStrengthMBadDreams } from '../mainskill/mainskills/charge-strengt
 import type { Pokemon } from './pokemon';
 
 export const DARKRAI: Pokemon = basePokemon({
-  name: 'DARKRAI',
   displayName: 'Darkrai',
   pokedexNumber: 491,
   specialty: 'all',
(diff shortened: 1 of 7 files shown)


3geek14:
When updating this PR to include the new Pokemon, I caught a bug in Minun! I think that makes 3 bugs this change has caught (previously Rufflet and Sneasel), which I think is good evidence that this change will help us to avoid bugs in the future!

common/src/utils/pokemon-utils/pokemon-constructors.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export type IngredientSetDefinition = {
+  ingredient0: IngredientSet[];
+  ingredient30: IngredientSet[];
+  ingredient60: IngredientSet[];
+};
+
+export type IngredientSpecification = IngredientDefinition | IngredientSetDefinition;
+
+function basePokemon(params: {
```
SleepAPI: Why are these called `baseXPokemon`? Are they not functions to create Pokemon objects? What is the base referring to?

I would expect something like:
```typescript
const CATERPIE = createBerryPokemon(...)
```
Or
```typescript
const CATERPIE = createPokemon('berry', ...)
```

common/src/utils/pokemon-utils/pokemon-constructors.ts
```diff
@@ -0,0 +1,193 @@
...
+  skillPercentage: number;
+  berry: Berry;
+  genders: GenderRatio;
+  carrySize: number;
+  previousEvolutions: number;
+  remainingEvolutions: number;
+  ingredients: IngredientSpecification;
+  skill: Mainskill;
+}): Pokemon {
+  const {
```
SleepAPI: This feels very verbose, which I understand has its own benefits, but we could also do something like:
```typescript
const { ingredients, specialty } = params;
const { ingredient0, ingredient30, ingredient60 } = getIngredientSets(ingredients, specialty);

return {
  ...params,
  ingredient0, ingredient30, ingredient60
};
```

common/src/utils/pokemon-utils/pokemon-constructors.ts
```diff
@@ -0,0 +1,193 @@
...
+    previousEvolutions,
+    remainingEvolutions,
+    ingredient0,
+    ingredient30,
+    ingredient60,
+    skill
+  };
+}
+
+export function baseAllPokemon(params: {
```
SleepAPI: This function out of context is quite confusing since "all" doesn't explicitly/intuitively link to specialty. Only way to really solve this would be either rename to something like `SpecialtyAll` or something like `createPokemon('all', ...)`

common/src/utils/pokemon-utils/pokemon-constructors.ts
SleepAPI: I see you already have this pattern in the other functions below too

SleepAPI (changes requested):

common/src/utils/pokemon-utils/pokemon-constructors.ts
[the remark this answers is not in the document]
3geek14: Added tests to confirm correct drops for Quaxly and Totodile. Quaxly has the unfortunate leek rounding, and Totodile is a non-ingredient mon and has only AB ingredients.

common/src/utils/pokemon-utils/pokemon-constructors.ts
SleepAPI: I usually call the remainder `rest`, but why do you do it in two steps?
I would probably just:
```typescript
 const { ingredients, specialty, ...rest } = params;
```

PR #790, opened 2025-08-19
feat(i18n): add translation files and update components with i18n keys

mabo-git-cham:
Add translation files and update components with i18n keys for multilingual support.
[ScreenShot.zip](https://github.com/user-attachments/files/21846705/ScreenShot.zip)


common/src/locales/en/files/nav.ts
```diff
@@ -0,0 +1,10 @@
+export default {
+  home: 'Home',
+  calculator: 'Calculator',
+  compare: 'Compare',
+  recipes: 'Recipes',
+  tierlist: 'Tierlist',
+  'game settings': 'Settings',
```
tindomerel8, line 7: Can we use camelCase instead of spaces for the translation string names, as below with NotFound? This would become gameSettings

common/src/locales/en/files/profilepage.ts
```diff
@@ -0,0 +1,7 @@
+export default {
+  profile: 'Profile',
+  aboutMe: 'About Me',
+  details: 'Details',
+  userID: 'UserID',
```
tindomerel8, line 5: ID is one word for the purpose of camelCasing, so this would be userId

PR #937, opened 2026-01-08
refactor: move strength calculation to server

webbywisp:


backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -141,6 +145,9 @@ export class MemberState {
   private level30IngredientAmount?: number;
   private level60IngredientAmount?: number;
 
+  // expert mode ingredient bonus — specialists roll an additional +1 at runtime
+  private expertIngredientSpecialistBonus = false;
```
tindomerel8, line 149: Worth future-proofing this for events, since we know events can cause pokemon to get +x ings for other reasons? Can always refactor it later though so feel free to disregard.

tindomerel8: I guess I don't know of other events that do a 50% chance though, maybe this is still unique

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -783,6 +801,46 @@ export class MemberState {
...
+    }
+    const pokemon = this.member.pokemonWithIngredients.pokemon;
+    if (!this.isExpertFavoredBerry(expertMode, pokemon.berry.name)) {
+      return;
+    }
+
+    // Expert mode ingredient modifier grants +1 ingredient per help for favored berries;
+    // ingredient specialists additionally roll for a further +1 at runtime.
+    this.expertIngredientSpecialistBonus = pokemon.specialty === 'ingredient';
+    this.level0IngredientAmount += 1;
```
tindomerel8, line 845: Does this always give them an extra lv 0 ing even if they didn't otherwise roll any of the lv 0 ing?

frontend/src/components/calculator/results/team-results.vue
```diff
@@ -199,59 +199,37 @@ export default defineComponent({
     },
     berryStrength() {
       const members = this.teamStore.getCurrentTeam.production?.members || []
-      const island = this.teamStore.getCurrentTeam.island
+      const weeklyFactor = timeWindowFactor('WEEK')
```
tindomerel8, line 202: Using this in a couple places, could just get it once. Minor cleanup, feel free to disregard.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
webbywisp: No, this happens statically in the constructor. This just increments the amount they will receive when they get an ingredient help by 1, it doesn't actually perform any helps. These are just the constant values that will be used if that makes sense

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
webbywisp: I think maybe keep this as is for now, this is kinda specific. Only applying to ingredient specialists, only applying 50% of the time. When events are properly introduced we will likely need to take a step back and look at it holistically again (did I use that right?)

frontend/src/components/calculator/results/team-results.vue
webbywisp: Good point, I'll fix this one

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
tindomerel8: Yep sounds good!

tindomerel8 (approved):

PR #945, opened 2026-01-31
fix(metronome): update list of blocked skills

3geek14:
Nitoyon ran an event to gather data from Metronome. From 1679 triggers, their data show a pretty flat distribution of how often each skill is rolled, with 24 skills each being rolled about 70 times. Some skills were never seen.

The results were shared with us on Discord at https://discord.com/channels/1300099710996058252/1328845974273134622/1465956055132667945


3geek14:
image

Some bars are a lot taller or shorter, but this is because the data represents what the user saw when collecting the proc, as opposed to what skill triggered. When Moonlight triggers, sometimes it looks identical to Charge Energy S. This data suggests the crit rate for Moonlight is about 55%, if we assume the two skills are equally likely to get rolled. Similarly, it suggests about an 11% rate for Berry Juice to give an item. Both of these results are based on a fairly small number of triggers, so I don't think we should update those odds based just on this, but I wanted to explain what might look like uneven odds of skills.

tindomerel8 (approved):

PR #969, opened 2026-04-07
Add Latias

3geek14:
lol I typoed the branch name oh well


3geek14:
image

3geek14:
In the following three images, note how many mons the skill detail claims get energy and helps per trigger, and note the total amount of energy and helps. With just Latias, only itself is healed. With a second mon, it heals both. With a full team, it's healing only two mons.

image

image

image

tindomerel8 (approved):

PR #1016, opened 2026-04-10
chore: update dependencies for common

Schanihbg:


common/src/utils/number-utils/number-utils.ts
```diff
@@ -1,3 +1,3 @@
-export function defaultZero(value?: number): number {
+export function defaultZero(value?: number | null): number {
```
3geek14: Both a `?` for the parameter and a `| null` for the type feels redundant. Is this intentional?

3geek14 (dismissed):

common/src/utils/number-utils/number-utils.ts
Schanihbg: Claude detected this


File | Issue | Fix
-- | -- | --
number-utils.ts:1 | Test passes null but signature only accepted undefined | Added null to the parameter type

3geek14: So maybe we shouldn't be passing `null` in our tests?

3geek14: We have a test for "it shall default to 0 for null", along with a couple other tests where it should default to 0. I don't think that means we want people passing `null` around anywhere, so I don't think we should update the signature. I'm fine leaving the test to show "see how robust this function is in the face of adversarial inputs".

Schanihbg: Hm, I'll revert back that change
I feel Claude didn't convince me enough

3geek14 (approved):

PR #1029, opened 2026-04-16
Guide page headers

tindomerel8:
Add `fullTitle` and `author` sections to frontmatter which build a new GuidesDocHeading for each page. This standardizes the H1 display to go along with the author, so guide authors should not include additional `#` H1s in the future.

Example of the new page header
image

Also, I got tired of Cursor writing curly quotes into code comments, so I banned it from using non-ASCII characters in comments and bulk updated those that had been added in a prior PR.


.cursor/rules/core-rules/code-comments-agent.mdc
```diff
@@ -20,6 +20,21 @@ alwaysApply: false
...
+## ASCII typography in comments (and docblocks)
+
+**Out of scope:** `**/*.md`. README, guides, and other Markdown may use curly quotes, em dashes, `×`, `→`, `…`, etc.
+
+**In scope:** Human-readable text in **`//`**, **`/* */`**, **`/** \*/`**, **`<!-- -->`\*\* (Vue), and SCSS/CSS comments. Keep that prose ASCII-friendly for tooling and diffs.
+
+- Use **straight apostrophe** `'` (U+0027) in words like `it's`, `Neroli's`, not curly single quotes
+- Use **straight double quotes** `"` (U+0022) in comments when you need quotes, not curly double quotes
+- Do **not** use **multiplication sign** `×` (U+00D7) for sizes in comments; write **`80x80`**, **`40x40`**, etc.
+- Do **not** use **em dash** `—` (U+2014) or **en dash** `–` (U+2013). Prefer **`-`** or **spaced hyphen** `-` for a break in a sentence
```
3geek14, line 32: Should it be
```
**spaced hyphen** ` - `
```
instead of
```
**spaced hyphen** `-`
```
for the second one? Feel free to disregard, if what you've got is working well enough.

.cursor/rules/core-rules/code-comments-agent.mdc
```diff
@@ -20,6 +20,21 @@ alwaysApply: false
...
+- Do **not** use **Unicode arrow** `→` (U+2192) in comments; use **`->`** or rephrase (e.g. maps to, becomes)
+- Use **three ASCII periods** `...` for ellipsis, not `…` (U+2026)
```
3geek14, line 34: But my custom keyboard firmware makes `…` so easy for me to type! I guess this means that we'll be able to distinguish between my comments and LLM comments based on which one gets used :P

guides/.vitepress/theme/utils/format-utils.ts
```diff
@@ -0,0 +1,44 @@
+// Convert an author display name to an avatar filename
+export function authorNameToAvatarSlug(displayName: string): string {
```
3geek14, line 2: These normalization rules look a bit complicated to me. If a new (non-technical) contributor wants to add their own avatar, are they going to have to ask us to do that?

guides/.vitepress/theme/style.scss
```diff
@@ -8,9 +8,14 @@
   --vp-custom-block-code-font-size: var(--text-code-size);
   font-family: var(--font-family-base);
 
-  .vp-doc {
+  .VPDoc {
     font-family: var(--font-family-prose);
 
+    // page H1 is `fullTitle` in doc-before; hide a stray markdown `#` if present
+    > h1:first-of-type {
+      display: none;
```
3geek14: If I write a doc like:
```md
# Intro

Blah blah

# Content

## Sub-heading

Yadda yadda
```
Will this hide my intro heading and keep my content heading? That feels weird to me. If the page title is h1, I'd expect my Intro and Content to both be bumped to h2, but I would prefer not to need to bump the headings myself in the markdown file.

guides/.vitepress/theme/vuetify.ts
```diff
@@ -6,11 +6,10 @@ import 'vuetify/styles';
 const guidesDarkTheme: ThemeDefinition = {
   dark: true,
   colors: {
-    background: '#191224',
-    primary: '#E63946',
-    secondary: '#5E5A7F',
-    surface: '#403D58',
-    accent: '#9A95C3',
+    background: '#191224', //neutral 900
```
3geek14, line 9: It's not clear to me what 900/500/700 are indicating. Am I part of the intended audience for the comments?

guides/tests/avatars.test.ts
3geek14, line 1: I think it's probably worth a test to ensure that the file names of avatars don't change when we try to normalize them. If I upload `3GEEK14.png`, forgetting that we normalize to lowercase rather than to uppercase, I'd appreciate a failing test telling me I've got it wrong, rather than needing to scour the code to see why my avatar isn't displaying.

guides/DEVELOPMENT.md
```diff
@@ -37,27 +37,32 @@ npm install
...
+Open the URL VitePress prints (default **http://localhost:5173/guides/** with this repo's config). The main Neroli's Lab app dev server uses a different port (see `frontend/vite.config.ts`); optional **`.env`** in `guides/` can set `VITE_MAIN_DEV_ORIGIN` so header links to `/`, `/calculator`, etc. point at that origin (see `.env.example`).
 
 ## Theme (SCSS)
 
-- **Colors:** `.vitepress/theme/tokens.scss` — CSS custom properties (`--color-*`, `--vp-c-*`).
-- **Typography:** `.vitepress/theme/typography.scss` — type scale and content rules.
-- **Layout / shell:** `.vitepress/theme/style.scss` — imports tokens + typography, slim nav, VitePress overrides.
-
-Custom Vue components should use the theme’s CSS variables (for example `var(--color-primary-500)`, `var(--tracking-tight)`).
+- **Colors:** `.vitepress/theme/tokens.scss` - CSS custom properties (`--color-*`, `--vp-c-*`).
```
3geek14, line 44: No change is showing up here…did there used to be non-breaking spaces in this section? 😬

guides/DEVELOPMENT.md
```diff
@@ -37,27 +37,32 @@ npm install
...
+- **Size limit:** each PNG must be ≤ **256 KiB**; enforced by **`tests/avatars.test.ts`**. **`npm run test`** catches oversize files locally; **CI** runs the same tests in `.github/workflows/build-test.yml` after **`npm run build`** for the guides job (tests are not chained into `npm run build`).
+
 ## Sidebar and content (technical summary)
 
-The sidebar is generated at build time from the **`content/`** tree. Folders become nested groups. YAML front matter supports **`title`** (nav label) and optional **`order`** (lower numbers sort earlier; ties use link order). Section folders may use **`index.md`** for the landing page and group title.
+The sidebar is generated at build time from the **`content/`** tree. Folders become nested groups. YAML front matter supports **`title`** (nav label), **`fullTitle`** (page H1), **`author`** (comma-separated list of names), and **`order`** (lower numbers sort earlier; ties use link order). Section folders may use **`index.md`** for the landing page and group title.
```
3geek14, line 59: "order (**optional;** lower numbers sort earlier; ties use link order)"

Is this a good place to keep a reference to it being optional?

guides/README.md
```diff
@@ -8,29 +8,31 @@ You do **not** need to be a developer to contribute. Most changes are plain **Ma
...
 
-You’ll need a GitHub account. If it's your first time, GitHub will walk you through forking and opening a PR from your fork.
+You'll need a GitHub account. If it's your first time, GitHub will walk you through forking and opening a PR from your fork.
 
 ## Adding a new page
 
 New guides are **Markdown files** under **`content/`**. The URL and sidebar follow the folder layout, so pick (or create) a folder that matches the topic.
 
-At the top of each file, YAML "frontmatter" sets how the page appears in the sidebar. Below that comes the Markdown body: **one** `#` heading for the page title, then `##` / `###` for subsections.
+At the top of each file, YAML "frontmatter" sets how the page appears in the sidebar. Below that comes the Markdown body. The main header `#` is reserved for the frontmatter, and `##` / `###` and other smaller headings can be used for subsections.
```
3geek14, line 17: If we auto-bump header numbers, this should get rewritten.

guides/README.md
```diff
@@ -40,16 +42,20 @@ Write your guide!
...
 
-If you are unsure where a new page should live, open an issue or ask in Discord—maintainers can help with structure.
+If you are unsure where a new page should live, open an issue or ask in Discord - maintainers can help with structure.
 
 ## Formatting with Markdown
 
-Use normal Markdown for headings, lists, links, and tables. The [Markdown Guide — basic syntax](https://www.markdownguide.org/basic-syntax/) explains common options. [VitePress Markdown extensions](https://vitepress.dev/guide/markdown) (tips, code blocks, etc.) work here too.
+Use normal Markdown for headings, lists, links, and tables. The [Markdown Guide - basic syntax](https://www.markdownguide.org/basic-syntax/) explains common options. [VitePress Markdown extensions](https://vitepress.dev/guide/markdown) (tips, code blocks, etc.) work here too.
+
+The top level `#` is defined in the frontmatter as `fullTitle`. **Do not** use additional `#` top-level headings. Use `##`, `###`, and smaller headings for sections inside the page.
```
3geek14, line 54: Goes away if we auto-bump headers

guides/README.md
```diff
@@ -40,16 +42,20 @@ Write your guide!
...
+
+## Author Avatars
 
-Every page should have **exactly one** top-level heading: a single `#` line (one H1). Use `##` and `###` for sections inside the page.
+You can optionally add a **square PNG** avatar in **`images/avatars/`** to be displayed automatically on pages you contribute to (when your name is listed in the `author` frontmatter). Name the file the same as your author name, with non-alphanumeric characters replaced with `-` (for example `jane-doe.png` for “Jane Doe” or `neroli-s-lab-team.png` for "Neroli's Lab Team").
```
3geek14: Still has curly quotes for Jane Doe.

Maybe worth including "ask for help if stuck", since the normalization is a bit complicated?

3geek14 (dismissed):
Most of my comments are nitpicks. I do want the extra test, but I'm approving so that you can add it in a fast-follow PR if you prefer, or in case I'm wrong and that test would be somehow a terrible idea.

guides/.vitepress/theme/style.scss
tindomerel8: I thought I deleted that rule. I'll remove it

guides/.vitepress/theme/utils/format-utils.ts
tindomerel8: It converts accented characters into their separate component parts then remove the accent parts leaving just the base character. It then replaces any number of non-[a-zA-Z0-9] characters with a single hyphen. I think this should be fairly intuitive to what people might expect to happen, or at least that's what I was going for. I explained it a bit in the README but I could add more examples to the README. If we need to help someone with the naming from time to time it should be okay imo.

Do you have ideas for a conversion that's simpler/more intuitive?

Examples in the test: 
```
expect(authorNameToAvatarSlug('Pat')).toBe('pat');
expect(authorNameToAvatarSlug('Jane Q. Contributor')).toBe('jane-q-contributor');
expect(authorNameToAvatarSlug('Example Author One')).toBe('example-author-one');
expect(authorNameToAvatarSlug('José')).toBe('jose');
```

guides/.vitepress/theme/vuetify.ts
tindomerel8: image

Our color palette in both Figma and SCSS use numbers for shades. It's a common pattern in the design world.

guides/DEVELOPMENT.md
tindomerel8: that one had curly quotes in "theme's" and an en-dash instead of a hyphen

guides/DEVELOPMENT.md
tindomerel8: I removed optional on purpose. I want to encourage people to consider the order of their pages, although it won't technically break without order.

guides/README.md
tindomerel8: We do not and cannot to the best of my knowledge. We have to rely on guide authors not using '#' (H1). This was already the case beforehand--we had to rely on them not putting more than one H1 in the page. If they do, it won't technically break anything, but it's not optimal for SEO or screen readers.

guides/README.md
tindomerel8: Do you have a proposed way for auto bumping headers? I have no idea how we'd do this without running some post-processor on the markdown and building the guides site from those instead.

guides/README.md
tindomerel8: They're allowed in human-facing pages, just not in code comments. I can add ask for help

guides/tests/avatars.test.ts
tindomerel8: We do run both the file names and the author names through the same normalization. 3GEEK14.png would match 3geek14 as an author name.

guides/.vitepress/theme/utils/format-utils.ts
3geek14: Trimming multiple hyphens down to one isn't a test case or in the README, right? `Pi's Bakery` goes to `pi-s-bakery`, but `Fishers' Bakery` (the one my whole family runs) does not go to `fishers--bakery`.

I think it's intuitive _enough_ to programmers (makes sense to normalize, and this is a perfectly reasonable way to do it), but I wouldn't expect it to be intuitive to non-programmers. And I think it's different from what Pandoc does to my section headings (I think `: ` becomes `--`), so I as a programmer would not assume before I read the code/comments/tests/README what it's doing.

guides/tests/avatars.test.ts
3geek14: sgtm

guides/DEVELOPMENT.md
3geek14: Oh, GitHub just isn't bothering to highlight which characters changed. Gotcha.

guides/README.md
3geek14: That was my thought, or perhaps doing post-processing on the compiled HTML, since maybe `<h1>` is less error-prone to find than `# `. But if there's no built-in support, then it's probably not worth doing.

guides/README.md
3geek14: I vaguely dislike treating `README.md` as more or less human-facing than other markdown files. I expect markdown files to pass through a renderer before being displayed to users, and I generally expect those renderers to handle curly quotes.

But really it was the inconsistency between `“Jane Doe”` and `"Neroli's Lab Team"` that bothered me. ;)

I appreciate the update to José to demonstrate diacritics.

3geek14 (dismissed):

guides/.vitepress/theme/utils/format-utils.ts
tindomerel8: Jane Q. Contributor test case does highlight the multiple non-alphanumeric characters becoming 1 hyphen. I can call it out in the readme

guides/README.md
tindomerel8: All markdown files except ones in an AI config (like the .cursor folder), I would consider intended to be readable to a non-programmer audience. All the guides are markdown, and this doc is what i'm pointing the aides to (presented nicely by GitHub's README previews when you navigate to the guides folder). 

But yeah I didn't like the inconsistency either; I just only mass-fixed the ones that were in code (non .md) files.

3geek14 (approved):

PR #1037, opened 2026-04-23
Add guides search

tindomerel8:


guides/.vitepress/lib/fill-local-search-preamble-map.ts
webbywisp, line 1: Do we have option for tests on these custom plugin things

webbywisp (approved):

guides/.vitepress/lib/fill-local-search-preamble-map.ts
tindomerel8: This particular file can't be tested without an additional dependency for DOM testing, but the other files do have test coverage. Merging for now with your approval but happy to revisit this later.

PR #1053, opened 2026-04-27
Refactor mainskill folders

3geek14:
When adding new mainskills, I'd like shortcuts to get to the folders where we implement them across common, backend, and frontend. Also, those folders are very densely packed, and I'd like subfolders for each base skill. This does that.


common/src/types/mainskill/mainskills/ingredient-magnet-s-present.test.ts
```diff
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest';
 import { IngredientMagnetS } from './ingredient-magnet-s';
-import { PresentIngredientMagnetS } from './present-ingredient-magnet-s';
+import { PresentIngredientMagnetS } from './ingredient-magnet-s-present';
```
tindomerel8: I notice you've renamed the file to move "Present" to the end, but left the exported skill name with the old style. Was this intentional? I see other skill names have been updated.

tindomerel8: Wait, no the other skill names have not been updated, but the effect names have.

tindomerel8: Image what we end up with is somewhat inconsistent

3geek14: Yeah, that's a mistake. I'll add that into `chore: rename skills to match in-games names`.

tindomerel8 (approved):

PR #1057, opened 2026-04-29
refactor: team skill targeting

3geek14:
Previously, energy skills targeted 1 or 5 members based on whether a `chanceToTargetLowestMembers` field was defined. And Nuzzle's extra skill activation rolls affected any mons that were healed by the same skill, which would not have worked if we gained a new skill that had the same skill activation effect but didn't restore energy.

Add fields to `SkillActivation` to handle each of these, and remove `chanceToTargetLowestMembers` from `UnitActivation`. At least so far, every skill that does multiple things to team members does the same things to each member it affects.

### DIFF
--- a/backend/src/services/simulation-service/team-simulator/skill-state/skill-state-types.ts
+++ b/backend/src/services/simulation-service/team-simulator/skill-state/skill-state-types.ts
@@ -1,12 +1,11 @@
-import type { Mainskill, MainskillUnit } from 'sleepapi-common';
+import type { Mainskill, MainskillTargeting, MainskillUnit } from 'sleepapi-common';
 
 export interface SelfActivationValue {
   regular: number;
   crit: number;
 }
-export interface TeamActivationValue extends SelfActivationValue {
-  chanceToTargetLowestMember?: number; // optional, if undefined this skill does not randomize between members
-}
+// TODO: Maybe replace both of these types with ActivationValue
+export type TeamActivationValue = SelfActivationValue;
 
 export interface UnitActivation {
   unit: MainskillUnit;
@@ -16,5 +15,6 @@ export interface UnitActivation {
 
 export interface SkillActivation {
   skill: Mainskill;
+  targeting?: MainskillTargeting;
   activations: UnitActivation[];
 }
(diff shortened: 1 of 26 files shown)


3geek14:
I found the bug! In `processTeamSkillHelpActivation`, I was returning early (and not giving any bonus skill helps) if `activation.unit` was anything other than `'helps'`…rather than `'skill helps'`. But the function was only called when it's `'skill helps'`, so the extra effect from Nuzzle was _never_ happening.

Comparison calc between prod and my local build now shows the same results:

image

image

backend/src/services/simulation-service/team-simulator/skill-state/skill-state-types.ts
```diff
@@ -1,12 +1,10 @@
...
+import type { Mainskill, MainskillTargeting, MainskillUnit } from 'sleepapi-common';
 
 export interface SelfActivationValue {
   regular: number;
   crit: number;
 }
-export interface TeamActivationValue extends SelfActivationValue {
-  chanceToTargetLowestMember?: number; // optional, if undefined this skill does not randomize between members
-}
+export type TeamActivationValue = SelfActivationValue;
```
tindomerel8, line 8: I guess we could probably rename this to just `ActivationValue` now and remove the extra type? Or do you foresee these diverging?

backend/src/services/simulation-service/team-simulator/team-simulator.test.ts
```diff
@@ -418,13 +422,164 @@ describe('recoverMemberEnergy', () => {
...
     expect(simulator.memberStates[0].energy).toBe(100);
     expect(simulator.memberStates[1].energy).toBe(50);
   });
+
+  it("shall recover some members' energy", () => {
+    const simulator = new TeamSimulator({
+      settings: mockSettings,
+      members: mockMembers.concat(mockMembers).concat(mockMembers),
+      iterations: 1
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
tindomerel8: Reasonable to disable this for the whole test file since this is frequently recurring here imo

tindomerel8 (dismissed):

backend/src/services/simulation-service/team-simulator/skill-state/skill-state-types.ts
3geek14: Probably, yeah. But I have a bunch of other skill refactors in my head, and I'd rather keep this here for now in case it becomes useful during those refactors. I can add a TODO to maybe delete it.

backend/src/services/simulation-service/team-simulator/team-simulator.test.ts
3geek14: Makes sense.

tindomerel8 (approved):

PR #1062, opened 2026-04-30
fix: cleanup skillValue in compare-strength.vue

3geek14:
* Use `defaultZero` in case `memberProduction.skillValue[activation]` is undefined. This can happen when a mon got 0 skill triggers, such as when it's sneaky snacking.
* Round once at the end, rather than once for each unit, for slightly higher accuracy. (Though, it's still mixing unrelated units, which is questionable.)
* Multiply by `timeWindowFactor` _before_ rounding, so that we don't divide by 3 (for 8H) after rounding.


tindomerel8 (approved):

tindomerel8 (changes requested):
oops, unapproving due to failing test

tindomerel8 (approved):

PR #1063, opened 2026-04-30
chore: remove row limit from compare tables

3geek14:
In Vue, `v-data-table` by default only allows up to 10 items to display at once. In the compare page, this happens to align with the most mons we allow the user to add. But if we ever decide to increase that, or if we clone those components for dynamic tier lists showing more than 10 mons, we'll need to remove the limit that Vue imposes. Setting `items-per-page="-1"` is how we remove the limit (which we've already done for the recipe table).

### DIFF
--- a/frontend/src/components/compare/compare-misc.vue
+++ b/frontend/src/components/compare/compare-misc.vue
@@ -3,7 +3,14 @@
   <v-row dense>
     <v-col cols="12">
       <v-card class="d-flex flex-column rounded-t-0 frosted-glass">
-        <v-data-table key="key" :items="members" :headers="headers" hide-default-footer class="bg-transparent">
+        <v-data-table
+          key="key"
+          :items="members"
+          :headers="headers"
+          hide-default-footer
+          class="bg-transparent"
+          items-per-page="-1"
+        >
           <template #item.member="{ item }">
             <div class="flex-center">
               <div style="overflow: hidden; width: 60px; height: 60px">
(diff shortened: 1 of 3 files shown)


frontend/src/components/admin/user-table/user-table.vue
```diff
@@ -11,6 +11,7 @@
       class="elevation-1"
       show-expand
       :loading="loading"
+      items-per-page="-1"
```
tindomerel8: We do actually use pagination on this table

tindomerel8 (changes requested):

frontend/src/components/admin/news-editor/announcements.vue
```diff
@@ -62,7 +62,13 @@
 
           <!-- History tab -->
           <v-window-item value="history">
-            <v-data-table :headers="historyHeaders" :items="newsHistory" :loading="isLoading" class="elevation-1">
+            <v-data-table
```
tindomerel8: This one too

frontend/src/components/admin/user-table/user-table.vue
3geek14: done

frontend/src/components/admin/news-editor/announcements.vue
3geek14: done

3geek14:
Now the only changes are in the compare pages.

tindomerel8 (approved):

PR #1064, opened 2026-04-30
feat: add custom emoji support to guides

tindomerel8:


guides/.vitepress/lib/guide-emojis.ts
```diff
@@ -0,0 +1,60 @@
...
+      category: parsed.category
+    };
+    if (byName.has(parsed.name)) {
+      collisions.add(parsed.name);
+    }
+    byName.set(parsed.name, entry);
+  }
+
+  if (collisions.size > 0) {
+    throw new Error(`Duplicate guide emoji basename(s): ${[...collisions].sort().join(', ')}`);
```
3geek14, line 43: I think it's worth a unit test that builds the map, makes sure it has more than zero emojis in it, and makes sure there are no collisions.

guides/content/index.md
```diff
@@ -29,4 +29,4 @@ Tooz is an editor for this project; we credit them for keeping guides organized
 
 ## Contributing
 
-This site is a community effort and contributions are welcome. If you've spotted an error, have data to add, or want to propose a new guide entirely, check out [this document](https://github.com/nerolis-lab/nerolis-lab/tree/main/guides/README.md) for info on how to contribute.
+This site is a community effort and contributions are welcome. If you've spotted an error, have data to add, or want to propose a new guide entirely, check out [this page](contributing) for info on how to contribute.
```
3geek14: "this page" isn't great link text. If someone is navigating with a screen reader, it's common to tab through links, and "this page; this page; that page; click here" being read aloud tells the user approximately nothing.

How about "check out our [contributing guide] for info"?

My father dislikes websites that tell him to do things, so he'd probably word it as "our [contributing guide] has info".

guides/scripts/sync-pokemon-emojis-from-frontend.mjs
3geek14, line 1: When does this get run? Is there a way to make it automatically happen when we add new images to the frontend? If not, it'd be great to put all the images into a shared project, and then to tell the guides emojis to ignore the versions of the pokemon images we don't want to use as emojis.

guides/.vitepress/lib/guide-emojis.ts
tindomerel8: I was going to, but this error serves the same purpose. Because this is run during the build step, it will fail the build on collisions (intentionally).

guides/scripts/sync-pokemon-emojis-from-frontend.mjs
tindomerel8: I set this up as a manual script for now, to avoid creating dependencies on frontend in the guides build process. When we refactor out that common UI component library, it might make sense to move shared public/ images there as well, then I would have no qualms about fully automating this process.

tindomerel8: It's run with npm run sync-pokemon-emojis

3geek14 (approved):

guides/content/index.md
tindomerel8: Pushed an update

PR #1065, opened 2026-05-01
Refactor: skill cleanups batch 1

3geek14:
Various cleanups in skill effect files. Each commit has one cleanup, so I recommend reviewing one commit at a time. If a reviewer gets through the first X commits and says "these are good so far, but I want a break", let me know where, and I'll split the PR at that point.

### DIFF
--- a/backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/cooking-assist-s/cooking-assist-s-bulk-up-effect.ts
+++ b/backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/cooking-assist-s/cooking-assist-s-bulk-up-effect.ts
@@ -1,32 +1,22 @@
 import type { SkillEffect } from '@src/services/simulation-service/team-simulator/skill-state/skill-effect.js';
 import type { SkillActivation } from '@src/services/simulation-service/team-simulator/skill-state/skill-state-types.js';
 import type { SkillState } from '@src/services/simulation-service/team-simulator/skill-state/skill-state.js';
-import {
-  CookingAssistSBulkUp,
-  emptyIngredientInventoryFloat,
-  flatToIngredientSet,
-  ingredient,
-  ingredientSetToFloatFlat
-} from 'sleepapi-common';
+import { CookingAssistSBulkUp, emptyIngredientInventoryFloat, flatToIngredientSet, ingredient } from 'sleepapi-common';
 
 export class CookingAssistSBulkUpEffect implements SkillEffect {
   activate(skillState: SkillState): SkillActivation {
     const skill = CookingAssistSBulkUp;
     const ingredientAmount = skillState.skillAmount(skill.activations.ingredients);
     const critAmount = skillState.skillAmount(skill.activations.critChance);
-    const magnetIngredients = emptyIngredientInventoryFloat().fill(
+    const flatIngredients = emptyIngredientInventoryFloat().fill(
       ingredientAmount / ingredient.TOTAL_NUMBER_OF_INGREDIENTS
     );
+    const ingredients = flatToIngredientSet(flatIngredients);
 
-    skillState.memberState.cookingState?.addIngredients(magnetIngredients);
+    skillState.memberState.cookingState?.addIngredients(flatIngredients);
+    skillState.memberState.addSkillProduce({ berries: [], ingredients });
+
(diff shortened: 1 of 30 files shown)


backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/cooking-assist-s/cooking-assist-s-bulk-up-effect.ts
```diff
@@ -1,32 +1,22 @@
...
     );
+    const ingredients = flatToIngredientSet(flatIngredients);
 
-    skillState.memberState.cookingState?.addIngredients(magnetIngredients);
-    skillState.memberState.cookingState?.addCritBonus(critAmount / 100);
+    skillState.memberState.cookingState?.addIngredients(flatIngredients);
+    skillState.memberState.addSkillProduce({ berries: [], ingredients });
 
-    skillState.memberState.skillProduce.ingredients = flatToIngredientSet(
-      ingredientSetToFloatFlat(skillState.memberState.skillProduce.ingredients)._mutateCombine(
```
tindomerel8, line 25: I notice this extra conversion has been removed in a number of places. Is it no longer needed? I don't see any change to flatToIngredientSet or see how the addSkillProduce method would make this unnecessary.

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/ingredient-magnet-s/ingredient-magnet-s-plus-effect.ts
```diff
@@ -24,7 +24,9 @@ export class IngredientMagnetSPlusEffect implements SkillEffect {
         member.skill.is(IngredientMagnetSPlus, CookingPowerUpSMinus)
       ).length === 0
         ? 0
-        : skillState.skillAmount(skill.activations.paired);
+        : skillState.skillAmount(skill.activations.paired, {
+            ingredient: skillState.memberState.member.pokemonWithIngredients.ingredientList[0].ingredient
```
tindomerel8, line 28: I notice this now handling the ingredients, but I don't see a similar removal of ingredient additions elsewhere. Was this being handled somewhere else that can now be cleaned up?

tindomerel8 (commented):
I've gotten through the first 4 commits so far and that's all I can do for right now.

Image

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/cooking-assist-s/cooking-assist-s-bulk-up-effect.ts
3geek14: The thing I removed was setting `memberState.skillProduce.ingredients` (an `IngredientSet[]`). The new value was computed by taking the old value, converting to a `Float32Array` ("float flat"), mapping from old to old+magnet, and converting back to an `IngredientSet[]`. Note that it used to convert both ways.

Instead, I call `addSkillProduce`, which calls `addToInventory` from `common`,, which operates on `IngredientSet[]`s without converting.

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/ingredient-magnet-s/ingredient-magnet-s-plus-effect.ts
3geek14: `skillAmount` in `skill-state.ts` (old line 224, new line 225) used to always pass this `ingredient` parameter to the `amount` function of the `ActivationType` it was given. Now it takes an optional `Partial<AmountParams>` argument and passes that along.

This is the one skill that used that parameter, so it's added here now that it's gone from `skillState.skillAmount`.

3geek14 (commented):
I've updated this PR to only include the 4 commits tindo has looked at so far.

tindomerel8 (approved):

tindomerel8:
The commits don't have the same SHAs as the ones I reviewed due to shuffling, but approving based on trust that they are the same as the ones I reviewed!

3geek14:
Yeah, they're different SHAs because I frequently rebase onto main. 😅 No merge conflicts from rebasing, so these changes are identical to what you reviewed.

PR #1069, opened 2026-05-02
Update early-progression.md

CyarsCaligo:


tindomerel8:
Thanks! Included your commit in https://github.com/nerolis-lab/nerolis-lab/pull/1071 and added a commit which fixes the linting, but can't merge this one as is. If this becomes a regular issue, we'll come up with a better way to address it.

PR #1070, opened 2026-05-03
feat: update common acronyms guide

tindomerel8:


guides/content/route-01/common-acronyms.md
```diff
@@ -1,81 +1,111 @@
...
-**EH** = Extra Helpful S  
-**HBoost** = Helper Boost¹  
-**IM/IngMag** = Ingredient Magnet S  
-**ID/IngDraw** = Ingredient Draw S  
-**BB** = Berry Burst  
-**TC** = Tasty Chance S  
-¹ In relation to Main Skills. Not to be confused with the Subskill `Helping Bonus`.
+| Term      | Definition                                              |
+| --------- | ------------------------------------------------------- |
+| CSS/CSM   | :chargestrength:&nbsp; Charge Strength S/M              |
```
3geek14, line 29: Why is `&nbsp;` needed in each of these? Should the template add this automatically?

3geek14 (approved):

guides/content/route-01/common-acronyms.md
tindomerel8: Cy thought it looked better with an extra space, and if you try to put "  " (two spaces), Prettier turns it into one space, so i used a non breaking space. I don't think it's generally applicable outside of this page.

PR #1071, opened 2026-05-03
Early game progression

tindomerel8:


guides/content/route-01/early-progression.md
```diff
@@ -11,19 +11,39 @@ The early game is a completely different ecosystem compared to the mid or late g
...
+
+That being said, we're not saying to "NEVER run Skill mons in the early stages." From a resource standpoint, we're just discouraging early-game players from actively hunting for Skill Mons as your Biscuit resources are better allocated towards hunting Berry and Ingredient mons. Though, if you _happen_ to find a Skill mon with long-term potential, it may be beneficial to invest in it.
+
+The single most important/impactful Skill to invest in is **Energy For Everyone (E4E)**. This allows your Pokémon to recover energy throughout the day which is disproportionately impactful from the mid game onwards. We do **NOT** recommend running temporary or univested E4E mons. The meta-defining E4E mons worth befriending are Gardevoir (Ralts), Pawmot (Pawmi), and Shuckle.
+
+Outside of E4E, there are 3 optional Main Skills that may be worth considering in your early game:
+
+- **Berry Burst (BB)**
+  - Treecko
```
3geek14: Why use basic forms here, vs. Gardevoir and Pawmot above?

3geek14 (approved):

guides/content/route-01/early-progression.md
tindomerel8: Used the evolution names in that earlier sentence because of the wording about those being the meta-defining options and to differentiate garde and gallade

guides/content/route-01/early-progression.md
```diff
@@ -11,19 +11,39 @@ The early game is a completely different ecosystem compared to the mid or late g
...
 
-You may have noticed a lack of mention for early game skill mons, and that’s because the necessity of an early game skill mon is highly debatable. **They require being fed 3-5 Main Skill Seeds to be viable to use so we rate them with higher standards, and they are not worth running without main seeds**. Despite this, it might be lucrative to invest in one to speed up your early game at the expense of needing to recover the main seeds used (reminder that with sleep points, premium players get an average of 1 main seed a month and f2p players get one every other month). If you are interested in what skill mons are worth investing in, visit: [Meta Analysis: Skill Specialists](../pokemon-evaluation/meta-skill.md)
+If you’re wondering which Berry mons to focus on, we recommend Pichu, Totodile, Cyndaquil, Chikorita, Spheal, Torchic, and Mudkip as they’re not only relatively common encounters on Greengrass Isle (GG), but also some of the best Berry mons for their respective islands.
```
jlau3: i think we should include Tyrunt now as well. Thoughts on Natu since it's very similar to Torchic? Or does the competition of so many good Lapis strength mons push it out?

I'm hoping in the future this will somehow not need to be manually maintained as it can maybe pull a list from the dynamic infographics? Or from the common folder itself (some kind of evaluation check for isOnGG, is5pip, > xyz strength)? Same idea for the other manual pokemon entries in these guides

guides/content/route-01/early-progression.md
```diff
@@ -11,19 +11,39 @@ The early game is a completely different ecosystem compared to the mid or late g
...
-- **Energy For Everyone:** This allows your pokemon to be healed up throughout the day. Healers are not particularly useful early game when your team is weak but an early one lets you take your time hunting for a really good one later on. These are not really recommended anymore due to acceleration to endgame. Mons of note: Igglybuff (do not catch Jigglypuff), Eevee (Sylveon), and Pawmi (capable of being a permanent healer with the appropriate stats)
+Sometimes you get unlucky with your morning spawns and you don’t encounter any worthwhile Berry mons. On these mornings, you can aim for Ingredient mons instead! While consistent cooking is usually a mid-game milestone and not typically a priority in the early game, you can still get a head-start by befriending Ingredient mons and looking for ones you will keep long-term.
```
jlau3: was this always so sparse? 😂 Might be a good idea to list out some common ing mons that are readily available on GG to mirror what we're doing for Berry mons. So Bellsprout, Croagunk, Geodude, gen 1 starters (should ABB squirtle be mentioned at all?), gen 9 starters, happiny, both woopers, pumpkaboo, cetoddle, and larvitar. Honorable mentions to Diglett for leeks at 30, Gastly for herbs/mushrooms, and slowpoke to unlock tails (only need 1)

Let me know if you want me to do the writeup if this is a good idea.

guides/content/route-01/early-progression.md
jlau3: I don't think Treecko is worth investing in early game due to how late BB scaling is and the fact we don't want to settle on a mon that potentially isn't being used until your team has built up some levels. Open to challenge that though

guides/content/route-01/early-progression.md
```diff
@@ -11,19 +11,39 @@ The early game is a completely different ecosystem compared to the mid or late g
...
+  - Treecko
+- **Charge Strength M (CSM)**
+  - Mareep
+  - Noibat
+  - Eevee (Espeon)
+- **Ingredient Magnet (Ing Mag)**
+  - Heracross
+  - Eevee (Vaporeon)
+
+Of note, _Premium Pass players_ have access to _1-2 MSS monthly_ compared to **F2P players** who have access to 1 MSS every **SECOND** month, on average. Premium players can be more generous with their MSS and can consider building multiple of the aforementioned Skill mons while F2P may struggle with the resources required to invest in one. Regardless of Premium or F2P status, _we recommend taking your time_ to hunt for ones that are good enough to be permanent additions to your team **BEFORE** spending the MSS.
```
jlau3: > who have access to 1 MSS every **SECOND** month, on average

Should mention that this is with our specific way of sleep point planning. People can absolutely just buy a main skill seed every month if that's their perogative

guides/content/route-01/early-progression.md
3geek14: With dish bonuses being so high, there's an argument to be made that tails for filler are detrimental to unlock, because one night prefer to get ingredients from magnet/tickets that can be put into strong meals. 

I think it makes sense to list some good ingredient mons. I don't know why cacao Blastoise would be excluded, since it's still an excellent mon, and it's arguably still the best cacao mon (because cacao dishes let you benefit from its milk, and rolling twice as many ABB Squirtle as AAA Wooper lets you hold out for an even better roll).

guides/content/route-01/early-progression.md
```diff
@@ -11,19 +11,39 @@ The early game is a completely different ecosystem compared to the mid or late g
...
+
+That being said, we're not saying to "NEVER run Skill mons in the early stages." From a resource standpoint, we're just discouraging early-game players from actively hunting for Skill Mons as your Biscuit resources are better allocated towards hunting Berry and Ingredient mons. Though, if you _happen_ to find a Skill mon with long-term potential, it may be beneficial to invest in it.
+
+The single most important/impactful Skill to invest in is **Energy For Everyone (E4E)**. This allows your Pokémon to recover energy throughout the day which is disproportionately impactful from the mid game onwards. We do **NOT** recommend running temporary or univested E4E mons. The meta-defining E4E mons worth befriending are Gardevoir (Ralts), Pawmot (Pawmi), and Shuckle.
+
+Outside of E4E, there are 3 optional Main Skills that may be worth considering in your early game:
+
+- **Berry Burst (BB)**
+  - Treecko
+- **Charge Strength M (CSM)**
```
jlau3: Do we no longer want to provide a reasoning as to why these 3 skills in particular are worth it over other main skills?

guides/content/route-01/early-progression.md
jlau3: Maybe we add a disclaimer for slowpoke then, I'd rather include all the nuances rather than have readers figure it out by omission. Something about tails being the strongest filler but filler is vastly overrated and can be actively detrimental if you're pursuing an ingredient magnet focused gameplay style.

And re: ABBstoise, yea I agree. I just know some ppl have reservations about ABB mons

guides/content/route-01/early-progression.md
tindomerel8: This isn't intended to be a full list, just one or two for each island. Taupe already has Cynda and Mudkip specified. All three of these sections would eventually link to the guides on meta mons of each type, like the guide in Discord does, but the links have been removed because those guides haven't been ported over yet.

guides/content/route-01/early-progression.md
tindomerel8: It was always this sparse because it links to the Ing mons guide, but we had to remove the link since that hasn't been ported over yet. I'd suggest keeping it sparse for now knowing that the link will be added when we can.

guides/content/route-01/early-progression.md
tindomerel8: Sure we can remove Treecko.

guides/content/route-01/early-progression.md
tindomerel8: Sure.

guides/content/route-01/early-progression.md
tindomerel8: Sure we can add it back. We shortened this section to avoid letting it meander too much and to keep it easy to follow, but that's reasonable.

guides/content/route-01/early-progression.md
jlau3: Ok, I'm just worried about some people that just take this at face value and don't look into other options so just a small addition saying this list is not exhaustive while we figure out the linking and the dynamic berry charts sounds good

jlau3 (dismissed):

guides/content/route-01/early-progression.md
tindomerel8: Sure. I wasn't sure whether we planned to go live without those links or not anymore. I'll add notes to each section

guides/content/route-01/early-progression.md
tindomerel8: Removed Treecko

guides/content/route-01/early-progression.md
tindomerel8: Notes added

guides/content/route-01/early-progression.md
tindomerel8: Note added

guides/content/route-01/early-progression.md
tindomerel8: Added

guides/content/route-01/early-progression.md
tindomerel8: Added a short description to each

jlau3 (approved):

PR #1082, opened 2026-05-04
feat: expand on contributing guide

tindomerel8:


guides/content/contributing/formatting.md
```diff
@@ -0,0 +1,189 @@
...
+
+To help readers distinguish between facts about game mechanics and subjective advice or recommendations, use the subjective info banner prior to subjective advice in a guide. Just add `<SubjectiveInfoBanner />` into your page and it will render like this:
+
+<SubjectiveInfoBanner />
+
+## Emphasis
+
+- **Bold** with `**double asterisks**`.
+- _Italic_ with `*single asterisks*` or `_underscores_`.
+- `Inline code` with backticks.
```
3geek14: Can you use \\\` to get backticks within inline code? I know this is always an obnoxious thing. Maybe ``<code v-pre>`backticks`</code>`` would work?

In GitHub comments, apparently I can use double backticks to start an inline code span that lets me display single backticks.

guides/content/contributing/formatting.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Lists
+
+### Ordered list
+
+1. First
+2. Second
```
3geek14, line 45: It also supports `1. Second`, right? Do we want to encourage that behaviour, as it makes adding/removing list items easier?

guides/content/contributing/using-github.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Using GitHub
+fullTitle: Contributing Guides with GitHub
+author: Tindo
+order: 60
+---
+
+This walkthrough is for **editing or adding guides entirely in the browser**, using GitHub’s interface. You do not need to install Git or editors on your computer.
```
3geek14: bunch curvy ’ in this file, and maybe in others

3geek14 (dismissed):

guides/content/contributing/formatting.md
tindomerel8: Prettier enforces numbering with 1.2.3. instead of 1.1.1. 

Considering whether to disable Prettier on these files because most contributors will be editing from their browsers without access to it, and it will just fail the builds until someone manually runs Prettier on them.

guides/content/contributing/using-github.md
tindomerel8: Gross. I only banned them in code comments but I hate that it's still doing this

guides/content/contributing/formatting.md
tindomerel8: Addressed

guides/content/contributing/using-github.md
tindomerel8: Replaced them all and updated Claude steering file

guides/content/contributing/formatting.md
3geek14: If we expect people (especially non-technical users) to edit in browsers without access to auto-formatting, then I think it's a good idea to disable linting for those files.

3geek14 (approved):

PR #1103, opened 2026-05-21
feat: updates for 3.5.0

3geek14:
Buffs to Wigglytuff and to meal energy recovery

* Wigglytuff buffed to 19.1 ing%, 22 inventory, and 2750 speed
* Meal energy recovery now gets bumped up by 1 for every 10 energy, starting at 2 recovered for 80.9 energy

To more accurately handle cases with fractional energy between 149 and 150, the `recoverMeal()` function in `MemberState` now computes a clamped recovery, and `getMealRecoveryAmount` no longer has a case where it returns 0.


backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -440,8 +440,10 @@ export class MemberState {
   }
 
   public recoverMeal() {
-    this.totalRecovery += getMealRecoveryAmount(this.currentEnergy);
-    this.currentEnergy += getMealRecoveryAmount(this.currentEnergy);
+    const recovered = getMealRecoveryAmount(this.currentEnergy);
+    const clampedRecovered = this.currentEnergy + recovered > 150 ? 150 - this.currentEnergy : recovered;
```
tindomerel8, line 444: We should probably clamp energy universally in a setter in the future so we don't have to manually do this here and forget to do this in other places. For now this is good though

tindomerel8 (approved):

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
3geek14: We clamp it in `recoverEnergy`, but that tracks wasted energy and attributes the wasted energy to the mon with the energy skill. But yeah, maybe there should only be one or two functions that manipulate energy, and they could take optional arguments for a mon with an energy skill.

PR #1172, opened 2026-07-14
chore: more package updates

tindomerel8:
Had to reformat some files with eslint --fix and prettier --write because prettier got stricter and changed its preferences for union types in an update.


common/src/utils/subskill-utils/subskill-utils.ts
3geek14, line 1: Huh…git blame says I left it like this? I really thought I did a multiline for this. I'm fine with either leaving it on one line the way the linter expects or changing it to multiple lines like this with the comment:
```
const numberOfElements =
  level < 10 ? 0 :
  level < 25 ? 1 :
  level < 50 ? 2 :
  level < 70 ? 3 :
  level < 80 ? 4 :
  5;
```

package.json
```diff
@@ -16,32 +16,32 @@
...
-    "eslint-plugin-vue": "^10.2.0",
-    "eslint-plugin-vuejs-accessibility": "^2.4.1",
-    "eslint-plugin-vuetify": "^2.5.2",
-    "globals": "^15.13.0",
-    "jiti": "^2.5.1",
-    "prettier": "^3.4.2",
-    "semantic-release": "^24.2.0",
-    "typescript": "^5.7.2",
-    "typescript-eslint": "^8.17.0"
+    "eslint": "^9.39.5",
```
3geek14, line 34: I think I found why so much needed to be re-linted. ;)

3geek14 (approved):

package.json
tindomerel8: Yeah that explains the linting changes, and prettier's update was the reason for the prettier failures.

PR #1206, opened 2026-07-27
feat: vue-i18n foundation with en/ja locales (replaces #790)

tindomerel8:
## Summary

Replaces #790 (stale since Aug 2025, conflicts with the current nav-bar/home-page structure) with a vue-i18n foundation built on top of `worktree-node-26-upgrade`. **The original translation commit from @masaki-fox is preserved via cherry-pick** (`git log` shows it as the first commit, authorship intact) rather than rewritten from scratch - everything after that is one squashed commit on top converting their content to JSON and adapting it to the current app structure.

- vue-i18n (Composition API) + `@intlify/unplugin-vue-i18n` for compile-time, type-safe messages
- Locale messages are **JSON** under `frontend/src/i18n/locales/{en,ja}.json` (one file per locale, namespaced internally) - per vue-i18n's own type-safe-resources guidance, Crowdin's best-supported format, and what `@intlify/eslint-plugin-vue-i18n`'s `localeDir` setting expects. The JSON values are ported directly from @masaki-fox's original `common/src/locales` TS files, not retranslated
- `numberFormats`/`datetimeFormats` wired into the same i18n instance, and the four duplicated `new Intl.NumberFormat/DateTimeFormat(navigator.language, ...)` call sites (`team-results.vue`, `cooking-results.vue`, `time-utils.ts`, `date-utils.ts`) now go through `n()`/`d()` - number/date formatting follows the selected UI language instead of the raw browser locale
- `language-store.ts` is the single source of truth for locale: `changeLanguage()` updates the store, `i18n.global.locale.value`, and `document.documentElement.lang` together; auto-detects from `navigator.language` on first load unless the user has manually chosen a language before (covered by a real, non-mocked test)
- Language switcher lives in Site Settings (existing `SettingsCard` pattern) rather than the nav drawer
- Nav drawer items translate via `t('nav.' + item.id)`, keyed off the stable `SiteNavItem.id` - not a `$route.name` string transform, which doesn't hold up against how many more routes main has today
- Addresses the outstanding review comments on #790: tierlist/userId key+value fixes tindomerel8 asked for, missing translations for Active/Inactive and "Supporter since" that webbywisp flagged, and removes the stray `diff.txt`/unused bun lockfiles/`cross-env` that were flagged
- New strings that were never part of #790's original Japanese work (nav.admin, the profile status strings) are English-only and fall back via `fallbackLocale` rather than being invented by me - they need a real translator or Crowdin
- Type safety on `t()` keys: `schema.d.ts` module-augments vue-i18n so `useI18n()` gets IDE-level key checking against the JSON schema, but that's autocomplete only - vue-i18n's typed `t()` has a permissive fallback overload so it never hard-fails on a bad key. `@intlify/eslint-plugin-vue-i18n`'s `no-missing-keys` rule is the actual CI-enforced check; see the code comment in `eslint.config.ts` for why it currently only reaches `.ts` files (a separate, pre-existing gap where none of this repo's vue-specific eslint rules run against `.vue` files via `eslint .`, unrelated to i18n and out of scope here)
- `SupportedLocale`/`SUPPORTED_LOCALES` now live in `common/src/locales` (re-exported from `frontend/src/i18n`) instead of being frontend-only, so `common/src/locales/en/pokemonNames.ts` - a separate, backend-shared dictionary - has a canonical locale list to agree with whenever it grows per-locale siblings, instead of two independently-drifting lists

## Out of scope (follow-ups)

- Translating every route's page title (only nav-drawer items are covered; app-bar title still shows the raw route name)
- Flag icons for the language switcher (dropped in favor of a simpler settings dropdown)
- Actual Crowdin project connection / CI sync workflow
- Real Japanese translations for the strings noted above as English-only/fallback
- Fixing `.vue` files being excluded from `eslint .`'s directory-walk entirely (pre-existing, affects every vue-specific rule already in the repo, not just the new i18n one - large blast radius, needs its own PR)
- Porting `common/src/locales/en/pokemonNames.ts` to vue-i18n: recommended against, not just deferred - it's consumed by the backend too, and vue-i18n's `t()` is a Vue-runtime API with no clean way to call it from Node

## Test plan

- [x] `npx vitest --run` in `frontend/` - 187 files, 1555 tests passing
- [x] `npx eslint frontend common` clean
- [x] `npm run type-check` clean
- [x] `vite build` succeeds, i18n messages code-split into their own chunk
- [x] Verified locally: language switch in Site Settings re-translates nav drawer, home page, and profile page live, and persists across navigation


3geek14:
We use a typescript module in `common` for names of Pokemon. The constructor asks for a key of the name type (e.g. `MR_MIME`) in order to pull the "localized" (only English so far[1]) display name (e.g. `Mr. Mime`). If the constructor is given a key that doesn't exist (e.g. `MX_MIME`), we catch the issue at compile-time, and we don't risk loading (and displaying) an empty name.

Will we have the same capacity for type-checking with this framework? Would it be in-scope to change `common`'s locales file(s) to the new framework?

[1]: because we only have English so far, I haven't tested that it draws the correct language. It very well might (read: probably) hard-code the English display name. Updating/implementing that (whichever is needed) can stay out-of-scope of this PR. Presumably the backend needs to be able to handle the frontend sending Pokémon in any language, which probably means never constructing any new Pokémon and never inspecting the display name from the backend.

PR #1213, opened 2026-07-29
fix: include all-specialists in every specialty filter chip

tindomerel8:
## Summary
- Darkrai's specialty is `'all'`, but the Pokemon search specialty filter did an exact match (`specialty === s`), so it never matched the Berry, Ingredient, or Skill chips - Darkrai showed under none of them.
- Filter now also passes when a Pokemon's specialty is `'all'`, consistent with how `common` already treats the all-specialist elsewhere (e.g. `stat-utils.ts`).

## Test plan
- [x] Added a test that fails without the fix and passes with it, for each of the Berry/Ingredient/Skill chips
- [x] `npx vitest --run` on PokemonSearch.test.ts (21 passed)
- [x] `npm run type-check` (frontend)

### DIFF
--- a/common/src/utils/pokemon-utils/pokemon-constructors.ts
+++ b/common/src/utils/pokemon-utils/pokemon-constructors.ts
@@ -4,7 +4,7 @@ import type { GenderRatio } from '../../types/gender';
 import type { Ingredient, IngredientSet } from '../../types/ingredient';
 import type { Mainskill } from '../../types/mainskill';
 import type { Pokemon, PokemonSpecialty } from '../../types/pokemon';
-import { calculatePityProcThreshold } from '../stat-utils/stat-utils';
+import { calculatePityProcThreshold, hasSpecialty } from '../stat-utils/stat-utils';
 import { evolvesFrom, evolvesInto } from './evolution-utils';
 
 export type IngredientDefinition = {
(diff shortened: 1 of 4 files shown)


frontend/src/components/pokemon-input/PokemonSearch.vue
```diff
@@ -286,7 +286,9 @@ const filteredPokemon: ComputedRef<PokemonWithPath[]> = computed(() => {
   }
 
   const specialtyFilter = (p: PokemonWithPath) =>
-    selectedSpecialties.value.length === 0 || selectedSpecialties.value.some((s) => p.pokemon.specialty === s)
+    selectedSpecialties.value.length === 0 ||
+    p.pokemon.specialty === 'all' ||
```
3geek14: Rather than adding this check every time we care about specialty, perhaps we should make a helper function for `hasSpecialty(mon: Pokemon, specialty: IForgetTheTypeName)`. Then All specialists can return true for that helper function.

common/src/utils/pokemon-utils/pokemon-constructors.ts
```diff
@@ -186,7 +186,7 @@ export function getIngredientSet(
 ): IngredientSet {
   const baseStrength = ingredientA.value;
   const levelFactor = level === 0 ? 1 : level === 30 ? 2.25 : 3.6;
-  const specialtyFactor = specialty === 'ingredient' || specialty === 'all' ? 2 : 1;
+  const specialtyFactor = hasSpecialty(specialty, 'ingredient') ? 2 : 1;
```
3geek14, line 189: I wish `hasSpecialty`'s first parameter was typed as `Pokemon`, so that we could generally avoid looking at `someMon.specialty` directly, but it looks like a handful of calls from this file really do want to pass a `PokemonSpecialty` instead of a `Pokemon`. I'll think about this in a follow-up so that we can merge the bugfix sooner.

3geek14 (approved):

PR #1214, opened 2026-07-30
fix: persist per-island berry settings and surface custom overrides

tindomerel8:
## Summary
- GG and GGEX (and every other island variant) now each remember their own last-saved berries/favorites/bonus independently, instead of a single team-wide island slot silently overwriting the others
- Unsaved edits also survive switching islands within the same open dialog, not just across saves
- Islands whose berries no longer match their in-game fixed set show a "Custom" badge (island cards, the calculator page icon, and the compare page's team selector) plus a "Default" button to revert
- Fixed the expert-mode favorite berry picker losing main-first ordering when switching between two expert islands with overlapping picks

## Test plan
- [x] `npx vitest --run` (full frontend suite, 1562 tests)
- [x] `npm run type-check`
- [x] `npx eslint` on all changed files


frontend/src/assets/common.scss
```diff
@@ -107,3 +110,16 @@
   display: inline-block;
   line-height: 0;
 }
+
+// same tag style as .expert-chip, for a setting that's been changed from its default
+span.custom-chip.text-small {
```
tindomerel8, line 98: I see the duplicated styles, but I plan to refactor this in a later pass. I want to make some larger typography changes first.

tindomerel8:
Pushed some changes. Saving now persists everything you did while in the modal, and flipping between islands no longer results in inconsistencies with the Custom tags

common/src/utils/island-utils/island-utils.test.ts
```diff
@@ -30,3 +30,35 @@ describe('getIsland', () => {
...
+  it('returns an empty array for Greengrass', () => {
+    expect(defaultIslandBerries(GREENGRASS.shortName)).toEqual([]);
+  });
+
+  it('returns an empty array for an expert island shortName', () => {
+    expect(defaultIslandBerries(GREENGRASS_EXPERT.shortName)).toEqual([]);
+  });
+});
+
+describe('hasCustomBerries', () => {
```
3geek14, line 48: Suggested additional tests, perhaps for a fast-follow PR rather than this one, so that this can get deployed sooner:

In 'is true when a base island differs from its default berries', add a line for the case where the berries are all of Cyan's plus another one:

```
  it('is true when a base island differs from its default berries', () => {
    expect(hasCustomBerries({ ...CYAN, areaBonus: 0, berries: [] })).toBe(true);
    expect(hasCustomBerries({ ...CYAN, areaBonus: 0, berries: [...CYAN.berries, berry.BELUE] })).toBe(true);
  });
```

In 'is false for Greengrass regardless of berries, since it has no fixed default', add a line for the base where there are no berries:

```
  it('is false for Greengrass regardless of berries, since it has no fixed default', () => {
    expect(hasCustomBerries({ ...GREENGRASS, areaBonus: 0, berries: [] })).toBe(false);
    expect(hasCustomBerries({ ...GREENGRASS, areaBonus: 0, berries: [CYAN.berries[0]] })).toBe(false);
  });
```

common/src/utils/island-utils/island-utils.ts
```diff
@@ -33,3 +33,20 @@ export function getIsland(nameOrBerries: IslandShortName | Berry[]): Area {
...
   return GREENGRASS;
 }
+
+export function defaultIslandBerries(shortName: IslandShortName): Berry[] {
+  const island = getIsland(shortName);
+  return island.expert ? [] : island.berries;
+}
+
+// Greengrass has no fixed default berries, so it's never "customized"
+export function hasCustomBerries(island: IslandInstance): boolean {
```
3geek14, line 43: This function looks weirdly formatted. Do we like inline if expressions? And does the return statement need parentheses? I guess it passed linting, so it's fine.

frontend/src/assets/common.scss
3geek14: Noted.

frontend/src/components/map/island-select.vue
```diff
@@ -56,14 +60,21 @@
                 <div class="flex-wrap title-row">
                   <span class="text-body island-name grow min-w-0">{{ filterIslandName(i.name) }}</span>
                   <span v-if="i.expert" class="expert-chip text-small">EX</span>
+                  <span
```
3geek14, line 63: Feels like a code smell that this block is being duplicated. Is one of them for mobile and the other for desktop?

3geek14 (approved):

frontend/src/components/map/island-select.vue
tindomerel8: Yes, there is a totally separate layout for desktop vs mobile here.

common/src/utils/island-utils/island-utils.ts
tindomerel8: Fast returns are about the only scenario where I don't mind inline ifs, personally (and the return statement does not need brackets or parens here). But I don't mind reformatting to full if blocks if you prefer.

3geek14: I like them for fast returns, but I'm used to everyone else always disliking them and configuring the linter to disallow them. 😅

PR #1216, opened 2026-07-31
Refactor/3geek14/skill cleanup batch 3

3geek14:
Cleanups in skill display components. Each commit can be reviewed separately. Happy to split this into multiple PRs after a partial review.

### DIFF
--- a/frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/berry-burst/berry-burst-details.vue
+++ b/frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/berry-burst/berry-burst-details.vue
@@ -1,6 +1,6 @@
 <template>
   <v-row no-gutters class="flex-center pb-1">
-    <v-col cols="auto" class="flex-center flex-nowrap mx-4">
+    <v-col cols="auto" class="flex-center flex-nowrap mx-2">
       <v-badge
         id="skillLevelBadge"
         :content="skillLevelBadgeText(effectiveSkillLevel, baseSkillLevel)"
--- a/frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-strength-s/charge-strength-s-range-details.vue
+++ b/frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-strength-s/charge-strength-s-range-details.vue
@@ -31,7 +31,7 @@
         </div>
         <div class="flex-left">
           <span class="font-weight-light text-body-2 text-no-wrap font-italic text-center"
-            >{{ localizeNumber(skillValuePerProc) }} avg.</span
+            >{{ localizeNumber(minStrengthPerProc) }}-{{ localizeNumber(maxStrengthPerProc) }}</span
           >
           <v-img src="/images/unit/strength.png" height="20" width="20" alt="strength" title="strength"></v-img>
         </div>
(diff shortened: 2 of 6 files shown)


3geek14:
The `displays the correct skill value per proc` tests don't look to me like they're testing any logic. They typically duplicate the logic from the function that calculates the per-proc amount and then also duplicate the formatting (e.g. `{value} avg.`) and then expect the real thing to match the freshly computed thing that was computed in the exact same way. So rather than updating each of those tests to do the new thing, I deleted them.

I probably should've also deleted the ones that check to make sure the area bonus is applied, because "correctly applies area bonus" should really be tested on the file that defines the function, rather than in these files calling it. But I left a couple of those in. The ones I removed are for skills that (in later commits) I changed the details component more. Partially because I was doing things out of order relative to the current commit list, and I didn't want to change the same test twice.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/berry-burst/berry-burst-details.vue
```diff
@@ -32,10 +30,26 @@
...
+            alt="berries"
+            title="berries"
+          ></v-img>
+        </div>
+        <div class="flex-left">
+          <span class="per-proc-amount font-weight-light text-body-2 text-no-wrap font-italic text-center mr-1"
+            >x{{ teamBerriesPerProc }}</span
           >
           <v-img src="/images/berries/berries.png" height="20" width="20" alt="berries" title="berries"></v-img>
+          <span class="font-weight-light text-body-2 text-no-wrap font-italic text-center"
```
tindomerel8: pl-1 (padding left 1) or pl-2 on this would help visually separate it from the berry. Same for Cress's berry burst and Latios

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-strength-s/charge-strength-s-range-details.vue
```diff
@@ -33,7 +31,7 @@
         </div>
         <div class="flex-left">
           <span class="font-weight-light text-body-2 text-no-wrap font-italic text-center"
-            >{{ localizeNumber(skillValuePerProc) }} avg.</span
+            >{{ localizeNumber(minStrengthPerProc) }}-{{ localizeNumber(maxStrengthPerProc) }}</span
```
tindomerel8, line 34: This is much longer than showing a single number, and so needs to be able to wrap on small screens. Here's 320px: Image

I can help with this if needed!

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/energy-for-everyone-s/lunar-blessing-energy-for-everyone-s-details.vue
```diff
@@ -33,17 +31,43 @@
         </div>
         <div class="flex-left">
           <span class="font-weight-light text-body-2 text-no-wrap font-italic text-center mr-1"
-            >x{{ skillValuePerProc }}</span
+            >x{{ energyPerProc }}</span
           >
           <v-img src="/images/unit/energy.png" height="20" width="20" alt="energy" title="energy"></v-img>
+          <span class="font-weight-light text-body-2 text-no-wrap font-italic text-center"
```
tindomerel8: Image and 

Image

Noting this to revisit after looking at cbex

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/berry-burst/berry-burst-details.vue
3geek14: Looks like most things have `mr-1` on the text before the icon, and Heal Pulse has `ml-1 mr-1` on the `x` between the icons, so I'll add that to all of these. And also to Helper Boost, in the following commit that deals with Helper Boost.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-strength-s/charge-strength-s-range-details.vue
3geek14: If it's as simple as adding U+200B (Zero Width Space) around the hyphen (which should maybe be an en dash, but shrug), then I can do that. But if it's more complicated, I would like help please. 🥺

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/energy-for-everyone-s/lunar-blessing-energy-for-everyone-s-details.vue
3geek14: :sderp:

tindomerel8 (approved):

PR #1230, opened 2026-08-05
feat: add Mew with placeholder skill

tindomerel8:
Implement the candy portion of Versatile and add Mew with placeholder skill% and placeholder text.

Still missing backend tests and frontend details component.


backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/versatile/versatile-effect.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  let versatileSEffect: VersatileSEffect;
+
+  beforeEach(() => {
+    memberState = mocks.memberState();
+    skillState = mocks.skillState(memberState);
+    versatileSEffect = new VersatileSEffect();
+  });
+
+  it('should include bonus candy when the roll succeeds', () => {
+    vimic(skillState, 'skillAmount', () => Versatile.candyAmount);
```
3geek14: Why is skill amount being vimic'd? This is just what this function should always return. I'm a bit more okay with vimic'ing the crit amount, but I'd rather set the mocked member state to have a skill level of 6+ to have non-zero crit.

frontend/public/images/avatar/happy/mew_happy_shiny.png
3geek14, line 1: Isn't shiny Mew blue? But also I assume it's shiny-locked in this game, so maybe that's why the shiny image looks the same as the normal image? If we're required to have shiny images to pass tests, I'd rather use real shiny images if we can. If we can't, maybe a future PR should block the shiny toggle in the UI for mons that are shiny locked (I think only Mew and Darkrai). If `common` knows that a mon is shiny-locked, the frontend tests can know to not look for shiny images.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/versatile/versatile-details.test.ts
```diff
@@ -0,0 +1,63 @@
+import MemberProductionSkill from '@/components/calculator/results/member-results/member-production-header/member-production-skill.vue'
+import type { MemberProductionExt } from '@/types/member/instanced'
+import { mocks } from '@/vitest'
+import type { VueWrapper } from '@vue/test-utils'
+import { flushPromises, mount } from '@vue/test-utils'
+import { MEW } from 'sleepapi-common'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
+
+const mockMember: MemberProductionExt = mocks.createMockMemberProductionExt({
+  member: mocks.createMockPokemon({ pokemon: MEW })
```
3geek14: I generally prefer to mock as much as feasible, which here would mean mocking deeper until we set the mainskill itself. I know this is copied from other tests that put in entire Pokemon, and I've been bad in the past, but maybe we can try to improve tests here?

common/src/types/mainskill/mainskills/versatile/versatile.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { MainskillTargeting } from '../../mainskill-targeting';
+
+export const Versatile = new (class extends Mainskill {
+  name = 'Versatile';
+  modifierName = 'Present';
+  RP = [1, 2, 3, 4, 5, 6, 7, 8];
+  candyAmount = 1;
+  bonusCandyAmount = [0, 0, 0, 0, 0, 1, 2, 3];
+  bonusCandyChance = 0.3;
+  image = 'metronome';
```
3geek14: Do we have access to the DNA icon for Versatile? It's only the wagging finger when Metronome is selected, but when I look at the skill info in the game, I see "About Versatile" with an icon matching what I see used on RaenonX.

Changing this will impact the frontend test.

3geek14 (changes requested):
Was waffling between approving or requesting changes. I'm going with requesting changes, specifically for the Versatile image.

frontend/public/images/avatar/happy/mew_happy_shiny.png
tindomerel8: It's shiny locked, and the solution for now is to make shiny look the same. I have a proper shiny lock implementation coming.

tindomerel8: See https://github.com/nerolis-lab/nerolis-lab/pull/1231, draft, not ready

common/src/types/mainskill/mainskills/versatile/versatile.ts
tindomerel8: I did look for the pink icon since I saw it in game, and I couldn't find it.

frontend/public/images/avatar/happy/mew_happy_shiny.png
3geek14: Heehee, same thoughts, but you were faster.

common/src/types/mainskill/mainskills/versatile/versatile.ts
3geek14: Rip. Maybe we can change it later.

3geek14 (dismissed):

3geek14 (approved):

PR #1231, opened 2026-08-05
fix: hide shiny toggle for shiny-locked pokemon

tindomerel8:
## Summary
- Add an optional `shinyLocked` flag to `Pokemon` and set it on Darkrai (mythicals/certain pokemon can't be shiny in-game)
- Hide the shiny toggle button in the pokemon editor for shiny-locked species, and no-op `toggleShiny()` as a guard
- Force `shiny` back to `false` when swapping an existing team member to a shiny-locked species via `createPokemonInstanceWithPreservedAttributes`

### DIFF
--- a/common/src/utils/pokemon-utils/pokemon-constructors.ts
+++ b/common/src/utils/pokemon-utils/pokemon-constructors.ts
@@ -35,6 +35,7 @@ function createPokemon(params: {
   remainingEvolutions: number;
   ingredients: IngredientSpecification;
   skill: Mainskill;
+  shinyLocked?: boolean;
 }): Pokemon {
   const { ingredients, ...otherParams } = params;
   const { name, specialty, frequency } = otherParams;
@@ -65,6 +66,7 @@ export function createAllSpecialist(params: {
   remainingEvolutions: number;
   ingredients: IngredientSpecification;
   skill: Mainskill;
+  shinyLocked?: boolean;
 }): Pokemon {
   return createPokemon({
     ...params,
@@ -85,6 +87,7 @@ export function createBerrySpecialist(params: {
   remainingEvolutions: number;
   ingredients: IngredientSpecification;
   skill: Mainskill;
+  shinyLocked?: boolean;
 }): Pokemon {
   return createPokemon({
     ...params,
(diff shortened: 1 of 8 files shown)


frontend/src/components/pokemon-input/pokemon-input.test.ts
```diff
@@ -99,6 +100,31 @@ describe('PokemonInput', () => {
...
+
+    await shinyButton.trigger('click')
+    expect(wrapper.vm.pokemonInstance.shiny).toBe(true)
+
+    await shinyButton.trigger('click')
+    expect(wrapper.vm.pokemonInstance.shiny).toBe(false)
+  })
+
+  it('hides the shiny toggle for shiny-locked pokemon', () => {
+    wrapper.vm.updatePokemon(mocks.createMockPokemon({ pokemon: DARKRAI }))
```
3geek14, line 115: It _probably_ won't happen to us, but other Pokemon games have a long history of releasing shiny forms of mons which were previously shiny locked. Is it worth importing a(nother) non-mocked Pokemon into this test file when we have the ability to mock deeper?

`pokemon: commonMocks.pokemon({ shinyLocked: true }})`

Since you create a mocked Darkrai multiple times, you could also extract this line to a constant somewhere.

frontend/src/services/utils/pokemon-instance-utils.test.ts
```diff
@@ -302,6 +303,28 @@ describe('createPokemonInstanceWithPreservedAttributes', () => {
     }
   })
 
+  it('should force shiny to false when the new Pokemon species is shiny-locked', () => {
+    const existingInstance: PokemonInstanceExt = mocks.createMockPokemon({
+      pokemon: PIKACHU,
```
3geek14, line 308: Similarly here, is it important to use Pikachu, Raichu, and Darkrai, or would two default mocks and a mocked shiny-locked mon work?

common/src/utils/pokemon-utils/pokemon-constructors.ts
3geek14, line 1: Evolution methods should probably pass along shiny-locked status. if Meltan is shiny-locked, then so too will be Melmetal.

3geek14 (approved):
Approving so I can merge and rebase Captain Pikachu onto this, since it's also shiny-locked, and I don't want to fake the images for it.

PR #1237, opened 2026-08-09
fix: correct CBEX island impact readout

tindomerel8:


frontend/src/components/calculator/results/member-results/island-impact/island-effects.ts
3geek14, line 1: Does it make sense for (much of?) this file to be moved into common? As is, this creates a second code location to update if CBEX is edited to have less severe nerfs. I'd rather the GGEX and CBEX numbers come from something in common that the island definitions can look at.

3geek14: I'll approve so we can land the bugfix, and I'll start working on the refactor.

3geek14 (approved):

PR #1245, opened 2026-08-12
simplify adding new Pokemon

3geek14:
Some cleanup to make it a little bit easier to add new Pokemon to the code.

* No longer need to add each new Pokemon to `INFERIOR_${foo}_SPECIALISTS` and `OPTIMAL_${foo}_SPECIALISTS`.
* No longer need to update backend snapshot tests when adding new mons doesn't change anything in the backend.


3geek14:
This is built on top of adding Hawlucha, but I'm pointing at `main` in order to run CI.

PR #1248, opened 2026-08-14
chore: create pokedex automatically

3geek14:
Currently, we manually create lists of "inferior" and "optimal" Pokemon, and we need to add to these lists every time we add a new mon. Then we combined the lists to get a full Pokedex. Refactoring lets us automatically add all new Pokemon to these lists, making one fewer chore when adding a new mon.

These lists are somewhat arbitrary, and were originally based on whether we wanted mons to appear in the tier lists and team finder. Pupitar was "good enough" to get included, in part because it has a different berry type (and a different favoured area) from Tyranitar. Swablu, despite having a different berry type and area fom Altaria, was not "good enough" to get included. Likely, going forward, we won't want to maintain both lists separately, but for now I'm keeping them for the sake of SleepAPI.

P.S. Also changed `Mainskill.hasUnit()` to take a parameter of type `MainskillUnit` rather than `string`, which revealed (and fixed) a bug that's been in SleepAPI probably since Dedenne came out. :D


3geek14:
In Discord, tindo wrote:

> I think if I was doing it from scratch, I probably would've gone the other way around and added Pokemon to a map when defining them rather than defining them each as a const with so many individual exports, and just put the names into a type to give strictly typed lookups for the map. But I haven't thought it through very much either so there might be an issue with that approach too

We already have a type for Pokemon names coming from the old l10n approach, which we use in Pokemon constructors: `keyof typeof pokemonNames`. I tried giving that an alias and implementing the Pokedex as `Map<PokemonName, Pokemon>`. It winds up being a pretty big refactor, due to all the places we currently accept a `string` for the species name, and needing to update those types or add new error handling. I do think a `Map` would be cleaner, but I'm abandoning this idea for now, and I'd like to stick with `Pokemon[]`.

I also looked into adding mons to the Pokedex when we create them, rather than in `pokedex.ts` afterwards. I was running into issues that I think involve initialisation orders, but using a `Pokemon[]` brings up another issue. When we create Pichu and Raichu, we edit the Pikachu object to include references to other forms in its evolution line. With a `Map`, I was able to do `pokedex.set('PIKACHU', updatedPikachu)` in the constructors for (pre)evolved mons. With an array, I'd need to `find` the old entry, remove it from the array, and add the new entry. This is pretty inelegant, and it's also slow, because removing stuff from arrays is slow.

For now, I think creating an array after importing all the mons is the cleanest approach.

3geek14:
Converting this to a draft to signify that it's blocked behind sunsetting tier lists, as this maybe breaks them by redefining "optimal".

PR #1279, opened 2026-09-02
chore: remove deprecated skillAmount

3geek14:
`skillAmount` was a `number` that summed the amount of each unit that a skill produced. Because it combined distinct units, it was deprecated when `skillValue` was introduced.

`skill-state.ts` still tracks the fields that were used to report `skillAmount`, but now nothing is using it, so those fields should be safe to remove.


frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-energy-s/charge-energy-s-details.vue
```diff
@@ -73,11 +73,15 @@ export default defineComponent({
...
       return this.memberWithProduction.member.skillLevel
     },
-    skillValuePerProc() {
+    strengthPerProc() {
       return ChargeEnergyS.activations.energy.amount({ skillLevel: this.effectiveSkillLevel })
     },
-    totalSkillValue() {
-      return compactNumber(this.memberWithProduction.production.skillAmount * this.timeWindowFactor)
+    totalStrength() {
+      const strengthSkillValue = this.memberWithProduction.production.skillValue['strength'] ?? {
```
tindomerel8: Are these called 'strength' universally, or should this be 'energy'?

tindomerel8: I see energy on another energy skill further down, so I assume this should be the same.

3geek14: This should be 'energy'. I'll poke through the charge energy skills. Do please lemme know if you find other examples where it's wrong.

3geek14: This one was also looking at `skillValue['strength']`, which meant it was showing 0. Whoops.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-energy-s/moonlight-charge-energy-s-details.test.ts
```diff
@@ -50,10 +50,4 @@ describe('Moonlight details', () => {
       MathUtils.round(mockMember.production.skillProcs * timeWindowFactor('24H'), 1).toString()
     )
   })
-
-  it('displays the correct total skill value', () => {
```
tindomerel8, line 54: Is there a reason for deleting these tests instead of updating them? I think they actually would've caught the miss.

3geek14: I forget why I deleted them, though I broadly oppose all the tests that find "whichever element has these generic classes" and think they should use element-specific classes/test-ids. And I didn't want to do that fixing during this PR.

I don't quite believe keeping this test would've caught the issue. Replacing `skillAmount` with `skillValue[someUnit]` in both places only would have caught the issue if I got `someUnit` correct in the test after getting it wrong in the UI, but I bet I'd be more likely to make the same mistake twice in a row.

I think my bigger issue with this test is that it's duplicating too much logic. `totalSkillValue` in the test looks like it's computed just about exactly the same way as `totalStrength()` was calculated in the UI, so the functionality being tested is "the number shows up in the element with the classes", rather than "the correct number shows up". And I put more trust in our ability to write `{{ totalSkillValue }}` in the right place than I place in our ability to list the right classes in the tests, especially if we ever adjust the styling.

3geek14: As to how I can (and now will) test this, I'm going to run it locally, load each distinct skill details component, and make sure there are no 0s.

tindomerel8: I agree on both counts: the selectors are bad and it should get a specific class to target, and that it duplicates too much logic. But I don't think the solution is to delete the test in this case; I think it's valuable to have a test that tests this (albeit without the duplicated logic and with more resilient selectors), and deleting it means it will likely never get fixed.

3geek14: > albeit without the duplicated logic

What sort of test do you think would be useful here? For context, here's what the test is doing at `main`:

We have a `const` mocked `MemberWithProduction`. We specify the `member: PokemonInstance` to be one that has the main skill we care about Before each test, we mount a skill details component for that mock. Then we duplicate the calculation of the total energy from the mock, by looking at `mockMember.production`.

But of course, this means we're looking at _default_ mocked production values. `createMockMemberProduction()`, in frontend/src/vitest/mocks/calculator/member-production.ts, gives the following produce:

* non-skill production:
  * 10 Belue berries at level 10
  * 10 Fancy Apples
  * 20 Honey
* skill production:
  * 0 berries and 0 ingredients
* total prudction:
  * matches non-skill production
* 5 skill triggers, providing:
  * 100 `skillAmount` (deprecated field)
  * 0 `amountToSelf` and 0 `amountToTeam` for every unit in `skillValue`
* 0s and empty arrays for everything else
* …overwritten with any parameters, so in theory we could specify what we want here

So when the test calculated `expectedValue` as `mockMember.production.skillAmount * timeWindowFactor('24H')`, that evaluates to `100 * 1`.

Maybe we want to mock deep enough to mock the `skillValue` map, and insert vaguely plausible values? I'll add a commit that adds a test like that to one of the skills. If you like it, I can add it elsewhere as well.

tindomerel8: This latest update looks good to me. There's value imo in testing that it's getting the value from the correct place on the mock Pokemon and compacting the number. If you think it's more trouble than it's worth, though, especially considering we hope to refactor these soon, I'll approve anyway for current state with most of them deleted.

tindomerel8 (approved):

PR #1287, opened 2026-09-03
Feat: projected snorlax rating

mjg-foundation:
Added a line to display the snorlax rank the expected strength falls in, given the selected area.


frontend/public/images/misc/rank-basic.webp
3geek14, line 1: I have a mild preference for png over webp, purely based on my recollection (as I review from my phone) that our other images are pngs.

What's the source for the images? The RaenonX team shares a bunch of assets with us, and I can send you ones for this.

frontend/src/components/calculator/results/team-results.vue
```diff
@@ -13,6 +13,13 @@
             </v-col>
           </v-row>
 
+          <v-row dense class="flex-center">
+            <v-col cols="auto" class="flex-center projected-rank">
+              <v-img :src="rankBallImage" :alt="`${projectedRank} Ball`" width="36" height="36" contain />
```
3geek14: Will the alt text be "Basic 1 Ball"? That feels wrong. Also, I think we generally include title text as well.

frontend/src/services/map-rank-service.test.ts
```diff
@@ -0,0 +1,13 @@
...
+import { describe, expect, it } from 'vitest'
+
+describe('projectedRankForStrength', () => {
+  it('uses the selected island thresholds', () => {
+    expect(projectedRankForStrength(100000, 'greengrass')).toBe('Ultra 2')
+    expect(projectedRankForStrength(100000, 'cyan')).toBe('Great 4')
+  })
+
+  it('uses the expert-map thresholds', () => {
+    expect(projectedRankForStrength(14780152, 'CBEX')).toBe('Master 20')
```
3geek14: I'd prefer for this test to match the earlier two and be a rank somewhere in the middle, rather than B1 or M20.

frontend/src/services/map-rank-service.ts
3geek14, line 1: This file feels like it should be in `common` rather than `frontend`. Rather than a lookup from short name to array, the arrays could be fields on the areas directly. 

Also, a bit of a nitpick, but calling it "_projected_ rank for strength" feels weird, since those strength numbers should always mean those ranks, without any need to make guesses. The projection is the guess about the final weekly strength number.

3geek14 (changes requested):

frontend/public/images/misc/rank-basic.webp
mjg-foundation: I curled them from reanonx 😅

tindomerel8: Webp is pretty solidly better than png today. We use pngs mostly out of outdated habits tbh. Switching to webp would be an actual upgrade.

tindomerel8: That said we should still use the original assets though, and if we decide to convert them to webp, do our own conversions.

mjg-foundation: pushed PNGs

common/src/types/island/island.ts
```diff
@@ -12,13 +12,15 @@ interface IslandBase {
 export interface Island extends IslandBase {
   berries: Berry[];
   expert: false;
+  rankThresholds: number[];
```
3geek14: Is there a reason to put this on `Island` and `ExpertIsland`, rather than on `IslandBase`? If we pass JSON blobs for `IslandBase` between the frontend and backend, it would be good to keep the extra data out of the network traffic, but I'm guessing we pass either `Area` or `shortName` instead, in which case that wouldn't matter.

common/src/utils/island-utils/rank-utils.ts
```diff
@@ -0,0 +1,11 @@
+import type { Area } from '../../types';
+
+const RANK_NAMES = ['Basic', 'Great', 'Ultra', 'Master'];
+
+export function rankForProjectedStrength(projectedStrength: number, area: Area): string {
```
3geek14: I'd still call this only `rankForStrength`. Whether the strength is projected or exact, this function doesn't care, right?

common/src/utils/island-utils/rank-utils.ts
```diff
@@ -0,0 +1,11 @@
...
+  const rankIndex = area.rankThresholds.findLastIndex((threshold) => threshold <= projectedStrength);
+  const titleIndex = Math.min(Math.floor(rankIndex / 5), RANK_NAMES.length - 1);
+  const rankNumber = titleIndex < 3 ? (rankIndex % 5) + 1 : rankIndex - 14;
```
3geek14: Rather than a conditional, I think you could do
```ts
const rankNumber = rankIndex - 5 * titleIndex + 1
```
because all but the final tier have 5 ranks, so you can always subtract 5 for each tier you've surpassed.

frontend/src/components/calculator/results/team-results.test.ts
```diff
@@ -50,6 +50,20 @@ describe('TeamResults', () => {
     expect(strengthSpan.text()).toBe('1,000')
     const totalStrength = wrapper.vm.totalStrengthString
     expect(totalStrength).toBe('1,000')
+    expect(wrapper.text()).toContain('Basic 1')
```
3geek14: Can these be more like the `strengthSpan` expectation, where we use `wrapper.find()` to get the element containing the text and the image, and then check the text and html of that element? I feel like "Poké Ball" might reasonably show up in multiple places on our Pokemon website at some point in the future, and I'd like this test to continue to look for the specific one we care about.

frontend/src/components/calculator/results/team-results.test.ts
```diff
@@ -50,6 +50,20 @@ describe('TeamResults', () => {
...
+    expect(wrapper.html()).toContain('alt="Poké Ball"')
+    expect(wrapper.html()).toContain('title="Poké Ball"')
+  })
+
+  it('uses canonical thresholds for legacy island instances', async () => {
```
3geek14: Rather than adding support for outdated team stores, we should add db migrations to update to having the new field. Migrations live in backend/src/database/migration/migrations and in frontend/src/stores/migration/migrations, depending on which databases need to be updated. The backend uses knex and the frontend uses pinia, so the migrations look different depending on which you're adding. :pensive:

frontend/src/components/calculator/results/team-results.vue
```diff
@@ -229,6 +244,25 @@ export default defineComponent({
     totalStrength() {
       return Math.floor(this.cookingStrength + this.berryStrength + this.skillStrength + this.stockpiledBerryStrength)
     },
+    projectedRank() {
+      const island = getIsland(this.teamStore.getCurrentTeam.island.shortName)
```
3geek14: And with database migrations, we shouldn't need `getIsland(someIsland.shortName)`.

frontend/src/stores/team/team-store.test.ts
3geek14: This snapshot is now 82 lines, 37 of which are added by this PR. Personally, I think inline snapshots for 20-line snapshots are kinda pushing the limit, so my preference would've been for these to be external snapshots even before your PR. Would you mind switching them while you're here? `expect(teamStore.$state).toMatchSnapshot()`, and then regenerating the snapshots will add them to an external file.

3geek14 (changes requested):
Looking pretty good! Aside from the database migrations, these comments are all pretty nitpicky. The code review style I learned at my first job (that actually had code reviews) is to include all the nitpicks, because sometimes authors overlook things. If an author responds with "I did it for such-and-such reason", often that's good enough. If you keep submitting PRs to this project, don't be shocked if you get a review from me that has a handful of comments about things I'd like to change and _also_ an approval: it means that I'm fine with you submitting, but I'd still like to hear responses to the comments I leave.

PR #1298, opened 2026-09-11
Mewtwo

mjg-foundation:
Adds mewtwo, psystrike, and berry zone.  
As of writing, mewtwo has placeholder stat values and ingredients list.  
Berry zone rotations are split into https://github.com/mjg-foundation/nerolis-lab/pull/2.


backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -387,6 +393,13 @@ export class MemberState {
 
   public addSkillProduce(produce: Produce) {
     this.skillProduce = CarrySizeUtils.addToInventory(this.skillProduce, produce);
+    const bonusBerries = this.berryZoneState.bonusBerries(produce.berries);
+    if (bonusBerries.length > 0) {
+      this.berryZoneSkillBonus = CarrySizeUtils.addToInventory(this.berryZoneSkillBonus, {
+        ingredients: [],
+        berries: bonusBerries
```
3geek14: It looks like berry zones are being implemented as granting extra berries, rather than as each berry itself being stronger. This means that everywhere we add berries, we also have to potentially add bonus berries from the zone, right? That feels easy to forget in the future.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -713,8 +729,17 @@ export class MemberState {
 
     const strength = new StrengthCalculator().calculateStrength({
       settings: this.settings,
-      produceWithoutSkill: totalHelpProduce,
-      produceFromSkill: totalSkillProduce,
+      produceWithoutSkill: {
+        ...totalHelpProduce,
+        berries: totalHelpProduce.berries.map((set) => ({
+          ...set,
+          amount: set.amount + this.berryZoneHelpBonus / iterations
```
3geek14: Does this mean that the number of berries reported will be inflated based on the berry zone? That feels like inaccurate reporting. I'd much prefer to report the accurate number of berries and indicate somewhere the average strength of those berries rather than pretending more berries were gathered to make the strength easier.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -108,6 +109,9 @@ export class MemberState {
   // summary
 
   // TODO: move to skill-state
+  public berryZoneState: BerryZoneState;
+  private berryZoneHelpBonus = 0;
+  private berryZoneSkillBonus: Produce = CarrySizeUtils.getEmptyInventory();
```
3geek14: It looks like berry zones are being implemented as granting extra berries, rather than as each berry itself being stronger. This feels like it will lead to confusion somewhere when tallying up how many berries have been gathered.

What if we instead added an field to `BerrySet` to track the berry zone bonus of those berries? Would that let us get rid of `berryZoneHelpBonus` and `berryZoneSkillBonus`?

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/berry-zone/psystrike-effect.ts
```diff
@@ -0,0 +1,23 @@
...
+export class PsystrikeEffect implements SkillEffect {
+  activate(skillState: SkillState): SkillActivation {
+    skillState.memberState.berryZoneState.addBonus(
+      berry.MAGO,
+      Psystrike.berryZoneAmount({ skillLevel: skillState.skillLevel }),
+      Psystrike.maximumBonus
+    );
+    return {
+      skill: Psystrike,
+      activations: [
```
3geek14: There should also be an activation with a new unit for the berry zone.

backend/src/services/simulation-service/team-simulator/skill-state/skill-state.ts
```diff
@@ -124,6 +127,7 @@ export class SkillState {
     this.rng = rng;
 
     this.skillEffects = new Map<Mainskill, SkillEffect>([
+      [Psystrike, new PsystrikeEffect()],
```
3geek14: Please keep these sorted alphabetically.

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/berry-zone/psystrike-effect.ts
```diff
@@ -0,0 +1,23 @@
+import type { SkillEffect } from '../../skill-effect.js';
+import type { SkillActivation } from '../../skill-state-types.js';
+import type { SkillState } from '../../skill-state.js';
+import { berry, Psystrike } from 'sleepapi-common';
+
+export class PsystrikeEffect implements SkillEffect {
```
3geek14: Please follow existing naming conventions for skills. This file should be `berry-zone-psystrike-effect.ts`, and the class should be named `BerryZonePsystrikeEffect`.

I do think that it would make sense to use the short names for skills in the file names, but I would prefer to keep things consistent for now. If you want to change things later, that can be its own commit to change all of them.

backend/src/services/simulation-service/team-simulator/berry-zone-state.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+describe('Psychic berry zone', () => {
+  it.each([
+    { level: 1, bonus: 0.6, triggers: 40 },
+    { level: 2, bonus: 0.8, triggers: 30 },
+    { level: 3, bonus: 1, triggers: 24 },
+    { level: 4, bonus: 1.2, triggers: 20 },
+    { level: 5, bonus: 1.6, triggers: 15 },
+    { level: 6, bonus: 2, triggers: 12 }
+  ])('requires $triggers activations at level $level to reach 24%', ({ level, bonus, triggers }) => {
```
3geek14: If the devs ever buff the skill and change the bonuses, this test would likely break. If the only code changes needed for such a buff would be in `common`, I don't want to add tests in `backend` that would need updates as well.

backend/src/services/simulation-service/team-simulator/berry-zone-state.test.ts
```diff
@@ -0,0 +1,76 @@
+import { describe, expect, it } from 'vitest';
+import { berry, MEWTWO, Psystrike } from 'sleepapi-common';
+import { BerryZoneState } from './berry-zone-state.js';
+import { mocks } from '@src/vitest/index.js';
+import { PsystrikeEffect } from './skill-state/skill-effects/berry-zone/psystrike-effect.js';
+
+describe('Psychic berry zone', () => {
```
3geek14: The file name says this is tests for `berry-zone-state.ts`, but here you say you're describing Psystrike. I think you want to split this file into two, and the tests for the state probably shouldn't reference the skill effect at all.

backend/src/services/simulation-service/team-simulator/berry-zone-state.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    expect(zone.bonusFraction(berry.MAGO)).toBe(0.24);
+    expect(zone.bonusFraction(berry.GREPA)).toBe(0);
+    zone.reset();
+    expect(zone.bonusPercentage(berry.MAGO)).toBe(0);
+  });
+
+  it('still grants Psystrike strength when the zone is already capped', () => {
+    const member = mocks.memberState({
+      member: mocks.teamMemberExt({
+        pokemonWithIngredients: { pokemon: MEWTWO, ingredientList: [MEWTWO.ingredient0[0]] }
```
3geek14: Does this care about being Mewtwo for anything other than the berry? I prefer deep mocking when possible.

backend/src/services/simulation-service/team-simulator/berry-zone-state.ts
```diff
@@ -0,0 +1,57 @@
+import { MathUtils, type Berry, type BerrySet } from 'sleepapi-common';
+
+interface ZoneBonus {
+  berryName: string;
+  percentage: number;
+  fraction: number;
+}
+
+/** Site-wide state: survives meals, sleep, and roster changes. */
```
3geek14: Simulation-wide, not site-wide, right? If I have multiple simulations, such as when comparing Mewtwos, the state shouldn't persist between those.

backend/src/services/simulation-service/team-simulator/berry-zone-state.ts
```diff
@@ -0,0 +1,57 @@
...
+export class BerryZoneState {
+  // Most teams have one zone. Its hot-path reads need no map lookup or division.
+  private firstBonus?: ZoneBonus;
+  private otherBonuses?: Map<string, ZoneBonus>;
+
+  private bonus(berry: Berry): ZoneBonus | undefined {
+    if (!this.firstBonus) return undefined;
+    return this.firstBonus.berryName === berry.name ? this.firstBonus : this.otherBonuses?.get(berry.name);
```
3geek14: Does the string comparison save time over a map lookup? How much does it save, to justify the added complexity in the code?

common/src/types/mainskill/mainskills/berry-zone/psystrike.ts
```diff
@@ -0,0 +1,20 @@
+import { Mainskill, type ActivationsType, type AmountParams } from '../../mainskill';
+
+export const Psystrike = new (class extends Mainskill {
```
3geek14: Please follow existing naming conventions here as well: `berry-zone-psystrike.ts` and `BerryZonePsystrike`.

common/src/types/mainskill/mainskills/berry-zone/psystrike.ts
```diff
@@ -0,0 +1,20 @@
...
+  strengthAmounts = [1408, 2002, 2762, 3813, 5264, 7274];
+  // Percentage points per activation, from the datamined infographic and its Lv.1 correction:
+  // https://www.reddit.com/r/PokemonSleep/comments/1wc455o/mewtwo_info_card_v380_update_my_mewtwo_event/
+  berryZoneAmounts = [0.6, 0.8, 1, 1.2, 1.6, 2];
+  maximumBonus = 24;
+  image = 'strength';
+  berryZoneAmount = this.leveledAmount(this.berryZoneAmounts);
+  description = (params: AmountParams) =>
+    `Increases Snorlax's Strength by ${this.activations.strength.amount(params)} and Mago Berry Strength by ${this.berryZoneAmount(params)}%, up to ${this.maximumBonus}%. The zone remains after switching Pokémon, until moving sites.`;
+  activations: ActivationsType = {
```
3geek14: There should also be a second activation here for the berry zone.

backend/src/services/simulation-service/team-simulator/skill-state/skill-effects/berry-zone/psystrike-effect.ts
```diff
@@ -0,0 +1,23 @@
...
+  activate(skillState: SkillState): SkillActivation {
+    skillState.memberState.berryZoneState.addBonus(
+      berry.MAGO,
+      Psystrike.berryZoneAmount({ skillLevel: skillState.skillLevel }),
```
3geek14: This probably wants to be replaced with a call to `skillState.skillAmount`.

common/src/types/mainskill/mainskills/metronome/metronome.ts
```diff
@@ -18,6 +19,7 @@ export const Metronome = new (class extends Mainskill {
     this,
     BerryBurstDisguise,
     ChargeStrengthMBadDreams,
+    Psystrike, // Keep the prototype out of random skill rolls until availability is confirmed.
```
3geek14: Please keep this list sorted alphabetically.

common/src/types/mainskill/mainskills/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './berry-burst';
+export * from './berry-zone/psystrike';
```
3geek14: Please add an index file to the new folder to keep this consistent.

common/src/types/pokemon/skill-pokemon.ts
```diff
@@ -1323,7 +1324,32 @@ export const PAWMOT: Pokemon = evolvedPokemon(PAWMO, {
   carrySize: 18
 });
 
+// Prototype only: speed, rates, ingredients, specialty and pity are provisional.
+export const MEWTWO: Pokemon = {
+  ...createSkillSpecialist({
```
3geek14: Using `{ ...createSkillSpecialist(` is a red flag here, and it doesn't match how any other Pokemon in this file are created. It looks like the only thing you're changing is the pity proc threshold…why do you want to circumvent the threshold that gets calculated by the constructor, especially with a provisional guess?

common/src/types/pokemon/skill-pokemon.ts
```diff
@@ -1323,7 +1324,32 @@ export const PAWMOT: Pokemon = evolvedPokemon(PAWMO, {
...
+    carrySize: 24,
+    previousEvolutions: 0,
+    remainingEvolutions: 0,
+    ingredients: {
+      a: MOOMOO_MILK,
+      b: SOOTHING_CACAO,
+      c: BEAN_SAUSAGE
+    },
+    skill: Psystrike,
+    shinyLocked: true
```
3geek14: Mewtwo will probably not be shiny locked, given shiny images for it have been found in the datamine.

common/src/types/pokemon/skill-pokemon.ts
```diff
@@ -1323,7 +1324,32 @@ export const PAWMOT: Pokemon = evolvedPokemon(PAWMO, {
   carrySize: 18
 });
 
+// Prototype only: speed, rates, ingredients, specialty and pity are provisional.
+export const MEWTWO: Pokemon = {
```
3geek14: Please keep this list sorted by Pokedex number.

common/src/types/pokemon/skill-pokemon.ts
```diff
@@ -1323,7 +1324,32 @@ export const PAWMOT: Pokemon = evolvedPokemon(PAWMO, {
...
+      c: BEAN_SAUSAGE
+    },
+    skill: Psystrike,
+    shinyLocked: true
+  }),
+  pityProcThreshold: 40
+};
+
 export const OPTIMAL_SKILL_SPECIALISTS: Pokemon[] = [
+  MEWTWO,
```
3geek14: This list as well gets sorted by Pokedex number.

backend/src/services/simulation-service/team-simulator/berry-zone-state.ts
mjg-foundation: switched to map lookup

backend/src/services/simulation-service/team-simulator/berry-zone-state.ts
mjg-foundation: corrected inaccurate comment

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
mjg-foundation: moved to a display that breaks out berry zone impact as a separate line item

common/src/types/berry/berry.ts
```diff
@@ -7,6 +7,8 @@ export interface BerrySet {
   amount: number;
   berry: Berry;
   level: number;
+  /** Average berry-zone strength bonus in percentage points, weighted by berry count. */
```
3geek14: Why average rather than exact? Is it easier to update a running weighted average than it would be to create a new `BerrySet` each time a mon collects a berry in a larger zone?

common/src/types/berry/berry.ts
```diff
@@ -7,6 +7,8 @@ export interface BerrySet {
...
+  berryZoneBonus?: number;
 }
 export interface BerrySetSimple {
```
3geek14, line 14: I think this type should likely be renamed to `BerrySetDto` (probably out of scope for your PR), as I expect it's a data transfer object type intended for passing abbreviated data between frontend and backend. It's a little suspicious to me that it doesn't also need `berryZoneBonus`. I'd expect that, somewhere, we pass a `BerrySetSimple` between services and then parse it as a `BerrySet`, and without the field, we'll always have an `undefined` bonus.

Looking more closely at where we use this type, it seems like it's only used for berry stockpiles at the starts of weeks. My guess is that, like mon levels, berry strength is based on the zone when the help happens, rather than the zone when the berry is collected, and it's technically possible to stockpile some buffed berries. But I'm fine with leaving out that functionality for now; I think stocking berries has mostly fallen out of style, and it would be an unusual strategy to run Mewtwo to make a big berry zone, then switch to a Johto cat, and then stockpile a bunch of buffed mago berries.

I think the only place where we calculate strength for this type is in frontend/src/components/calculator/results/team-results.vue (as apparently we never send stockpiled berries into the simulation), and that should probably be updated (also out of scope for your PR) to call strength service. So you probably don't have to update the DTO type in your PR, but could you please at least add a TODO comment to synchronize the types?

common/src/types/mainskill/mainskills/berry-zone/berry-zone-psystrike.ts
```diff
@@ -0,0 +1,18 @@
+import { Mainskill, type ActivationsType, type AmountParams } from '../../mainskill';
+
+export const BerryZonePsystrike = new (class extends Mainskill {
+  name = 'Psystrike (Berry Zone)';
+  // Placeholder RP values; replace when verified RP data is available.
+  RP = [1408, 2002, 2762, 3813, 5264, 7274];
+  strengthAmounts = [1408, 2002, 2762, 3813, 5264, 7274];
+  berryZoneAmounts = [0.6, 0.8, 1, 1.2, 1.6, 2];
+  maximumBonus = 24;
+  // TODO: Replace the 72x72 icon with a 256x256 version before merging.
```
3geek14: I'd remove the "before merging" from this TODO. :P It's a good TODO to do once we have the right image, but I'm not worried about merging and then doing a follow-up afterwards.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/psystrike-berry-zone/psystrike-berry-zone-details.vue
```diff
@@ -0,0 +1,99 @@
...
+    return { teamStore, skillLevelBadgeText, mainskillImage, berryImage, berry, compactNumber, localizeNumber }
+  },
+  computed: {
+    effectiveSkillLevel() {
+      return this.memberWithProduction.production.skillLevel
+    },
+    baseSkillLevel() {
+      return this.memberWithProduction.member.skillLevel
+    },
+    skillValuePerProc() {
```
3geek14: `strengthPerProc()` instead, because part of the value of the skill is the berry zone. There's a refactor in review that would've removed whatever you copied this name from. Same with `totalStrength()` below.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/psystrike-berry-zone/psystrike-berry-zone-details.vue
```diff
@@ -0,0 +1,99 @@
...
+            src="/images/misc/skillproc.png"
+            height="24"
+            width="24"
+            alt="skill activations"
+            title="skill activations"
+          ></v-img>
+        </div>
+        <div class="flex-left">
+          <span class="font-weight-light text-body-2 text-no-wrap font-italic text-center mr-1"
+            >x{{ localizeNumber(skillValuePerProc) }}</span
```
3geek14: Check out Delibird's skill for an example. I'd like to show both the strength per proc and the berry zone per proc info here.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/psystrike-berry-zone/psystrike-berry-zone-details.vue
```diff
@@ -0,0 +1,99 @@
...
+      return this.memberWithProduction.member.skillLevel
+    },
+    skillValuePerProc() {
+      const rawAmount = BerryZonePsystrike.activations.strength.amount({ skillLevel: this.effectiveSkillLevel })
+      return applyAreaBonus(rawAmount, this.teamStore.getCurrentTeam.island.areaBonus)
+    },
+    totalSkillValue() {
+      return compactNumber(this.memberWithProduction.production.strength.skill.total * this.timeWindowFactor, 'floor')
+    },
+    totalBerryZoneBonus() {
```
3geek14, line 99: I think https://github.com/nerolis-lab/nerolis-lab/blob/a6be84460b6ae11abdf74427148fddf15db0f6e2/frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/charge-energy-s/charge-energy-s-details.vue#L82 has a nicer pattern for this.

3geek14 (changes requested):
Partial review as I've got to head out for a bit.

common/src/utils/berry-utils/berry-utils.ts
```diff
@@ -78,3 +78,16 @@ export function uniqueMembersWithBerry(params: { berry: Berry; members: Pokemon[
   );
   return count;
 }
+
+/** Merge matching berry types and levels without changing their total strength or count. */
+export function mergeBerrySets(first: BerrySet, second: BerrySet): BerrySet {
```
3geek14: If we don't call this function a ton, then it's probably worth adding some error handling to make sure we never mix sets with differing berries/levels.

common/src/utils/carry-size-utils/carry-size-utils.ts
```diff
@@ -16,11 +17,7 @@ class CarrySizeUtilsImpl {
         );
 
         if (index !== -1) {
-          newInventory.berries[index] = {
-            berry: newInventory.berries[index].berry,
-            amount: newInventory.berries[index].amount + produceBerrySet.amount,
-            level: newInventory.berries[index].level
-          };
+          newInventory.berries[index] = mergeBerrySets(newInventory.berries[index], produceBerrySet);
```
3geek14: Why do you want to merge them here rather than adding a comparison to `findIndex` on lines 15–17? My preference is to track more details during the simulations, and only combine information at the end when presenting it to users. That way, if we want to provide more details in the future, we don't have to plumb more tracking information through the simulation.

frontend/src/components/calculator/results/member-results/member-production-header/member-production-berry.vue
```diff
@@ -48,9 +44,16 @@ export default defineComponent({
     return { teamStore, MathUtils, berryImage, isMobile }
   },
   computed: {
+    berryCountLabel() {
+      const amount = this.memberWithProduction.production.produceWithoutSkill.berries.at(0)?.amount ?? 0
```
3geek14: This is where I would prefer to merge berry sets. Probably one function that merges all the berry sets in `this.memberWithProduction.production.produceWithoutSkill.berries` and then use the result of that in both `berryCountLabel()` and `averageZoneBonus()`.

Then you'd probably want the merge function to take a `BerrySet[]` rather than just two. And perhaps add some error handling in case the array given to the merge is somehow empty. Perhaps force callers to pass in a berry type and a level, and return a berry set with `amount: 0` if given an empty array.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -155,12 +158,19 @@ export class MemberState {
     team: TeamMember[];
     settings: TeamSettings;
     cookingState: CookingState | undefined;
+    berryZoneState?: BerryZoneState;
```
3geek14: Do we ever create a `MemberState` without having access to an existing `BerryZoneState`? I'm surprised to see this parameter be optional.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -155,12 +158,19 @@ export class MemberState {
...
     iterations?: number;
     rng?: PreGeneratedRandom;
   }) {
     const { member, team, settings, cookingState, iterations = 1, rng } = params;
 
     this.rng = rng || createPreGeneratedRandom();
+    this.berryZoneState = params.berryZoneState ?? new BerryZoneState();
```
3geek14: You can add `berryZoneState` into the assignment on line 165 rather than using `params.` on this line.

backend/src/services/simulation-service/team-simulator/member-state/member-state.ts
```diff
@@ -505,7 +518,10 @@ export class MemberState {
     const totalDropAmount = berryAmount + ingredient0Amount + ingredient30Amount + ingredient60Amount;
 
     if (ingredientId === undefined) {
-      this.totalBerryProduction += totalDropAmount;
+      this.totalBerryProduction = mergeBerrySets(
```
3geek14: If the `totalBerryProduction` were an array instead of a single `BerrySet`, you wouldn't have to merge them here, and you could track how much each berry is worth.

Also, now that this one-line statement has turned into 4 lines, it might be worth a helper function to use here and in `attemptSneakySnackingHelp` and `attemptNightHelp`.

3geek14 (changes requested):
Finished reviewing this version. Looking like we'll be done soon!

common/src/types/berry/berry.ts
mjg-foundation: added a TODO, and created #1308

frontend/src/components/calculator/results/member-results/member-production-header/member-production-berry.vue
```diff
@@ -48,9 +44,22 @@ export default defineComponent({
     return { teamStore, MathUtils, berryImage, isMobile }
   },
   computed: {
+    mergedBerries() {
+      const { member, production } = this.memberWithProduction
+      const sets = production.produceWithoutSkill.berries
+      const first = sets.at(0)
+      return mergeBerrySets(sets, first?.berry ?? member.pokemon.berry, first?.level ?? member.level)
```
3geek14: We always expect `first.berry` and `first.level` to match `member.pokemon.berry` and `member.level`, right? If that's the case, I'd leave out the `first?.foo ??` from both and never define `first`. What desync are you worried about to justify the extra work?

common/src/types/mainskill/mainskills/berry-zone/berry-zone-psystrike.ts
```diff
@@ -0,0 +1,17 @@
+import { Mainskill, type ActivationsType, type AmountParams } from '../../mainskill';
+
+export const BerryZonePsystrike = new (class extends Mainskill {
+  name = 'Psystrike (Berry Zone)';
+  RP = [2450, 3383, 4733, 6551, 8994, 12028];
+  strengthAmounts = [1408, 2002, 2762, 3813, 5264, 7274];
+  berryZoneAmounts = [0.6, 0.8, 1, 1.2, 1.6, 2];
+  maximumBonus = 24;
+  // TODO: Replace the 72x72 icon with a 256x256 version.
+  image = 'psystrike';
```
3geek14: `image = 'berryzone';` for the image tindo added, and remove the comment

common/src/types/pokemon/skill-pokemon.ts
```diff
@@ -373,6 +374,26 @@ export const FLAREON: Pokemon = evolvedPokemon(EEVEE, {
   skill: CookingPowerUpS
 });
 
+// Prototype only: speed, rates, ingredients, and specialty are provisional.
```
3geek14: no longer provisional, so this comment can go away

frontend/src/components/calculator/results/member-results/member-production-header/member-production-skill-details/psystrike-berry-zone/psystrike-berry-zone-details.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    wrapper = mount(MemberProductionSkill, { props: { memberWithProduction: member } })
+    await flushPromises()
+    await vi.dynamicImportSettled()
+  })
+
+  afterEach(() => wrapper.unmount())
+
+  it('loads Psystrike details with the new icon and skill level badge', () => {
+    expect(wrapper.findComponent({ name: 'PsystrikeBerryZoneDetails' }).exists()).toBe(true)
+    expect(wrapper.find('img').attributes('src')).toBe('/images/mainskill/psystrike.png')
```
3geek14: `berryzone` image

frontend/public/images/mainskill/psystrike.png
3geek14, line 1: this file is no longer needed

frontend/src/services/utils/image-utils.ts
```diff
@@ -14,6 +14,8 @@ export function mainskillUnitImage(unit: MainskillUnit) {
   switch (unit) {
     case 'berries':
       return '/images/unit/berry.png'
+    case 'berry zone':
+      return '/images/mainskill/psystrike.png'
```
3geek14: `/images/unit/berryzone.png`

3geek14 (changes requested):
pretty much done, aside from using the images tindo has added

3geek14 (approved):

PR #1299, opened 2026-09-11
Feat: Raptor BFS Index

mjg-foundation:
Added a live Raptor BFS Index page to the guides site, with toggles for NFE pokemon and applying the IFM adjustment; and quick search.


3geek14 (changes requested):
I think this would be better suited to `guides` than to `frontend`. We're probably removing the existing cooking tier list and we'd like to replace it with an assortment of new tier lists, which will live in the guides section.

